PR #98, opened 2014-12-18
Add support for UUID version 1 to UUIDFactory

erikryverling:
Also improved documentation and minor cleanup.

The UUID1 implementation is standardized and compatible with other implementations, such as [python's uuid.uuid()](https://docs.python.org/2/library/uuid.html#uuid.uuid1). 

@adamvoncorswant @fredrikbackstrom Please have a look.


izettle-java/src/main/java/com/izettle/java/uuid/uuid1/UUID1Utils.java
```diff
@@ -0,0 +1,22 @@
...
+	private UUID1Utils() {
+	}
+
+	/**
+	 * Extracts a UUID timestamp and returns it as a Date.
+	 * @param uuid1 The UUID to extract from.
+	 * @return The extracted date.
+	 */
+	public static Date getDateFromUUID1(UUID uuid1) {
+		return new Date((uuid1.timestamp() - CLOCK_OFFSET) / CLOCK_MULTIPLIER);
```
erikryverling: Right

izettle-java/pom.xml
```diff
@@ -1,5 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
```
erikryverling: Yeah, thanks

izettle-java/src/main/java/com/izettle/java/uuid/uuid1/UUID1Generator.java
```diff
@@ -0,0 +1,52 @@
+package com.izettle.java.uuid.uuid1;
```
erikryverling: Yeah, maybe. I'll change it.

izettle-java/src/main/java/com/izettle/java/uuid/UUIDFactory.java
```diff
@@ -0,0 +1,113 @@
+package com.izettle.java.uuid;
```
erikryverling: Well, I think it makes sense since we now have three classes that all regards UUID, and generally I think having a descriptive and logic package structure is something that is a bit lacking throughout our code base.

Regarding backwards compatibility I get you point that it will be a bit of work for the client when updating (although in reality it's usually a matter of clicking on the now moved class and just change the import statement(s)).

I think there's a danger of refraining from sound refactorings due to fear of breaking backwards compatibility, but if you tell me that this change will have a severe impact on the users of the toolbox, I can of course leave it in the original package. 

Do we have a policy of how we increment the version numbers? If we're using something like semantic versioning, we should make sure to increment the major version to indicate that there will be an API change,

izettle-java/src/main/java/com/izettle/java/uuid/uuid1/UUID1Generator.java
```diff
@@ -0,0 +1,52 @@
...
+
+/**
+ * Generates a new UUID version 1. The UUID does NOT contain any host information, but only time information.
+ *
+ * This class is based on the java-uuid-generator project.
+ * @see <a href="https://github.com/cowtowncoder/java-uuid-generator">
+ *     https://github.com/cowtowncoder/java-uuid-generator
+ *     </a>
+ */
+public final class UUID1Generator {
```
erikryverling: Ah... naming. The worst part of being a developer. My reasoning for calling it `UUID1Generator` is that it's only purpose is to generate a UUID version 1, where the `UUIDFactory` both generates and Base64 encodes. But I can change it if you think it's more clear.

erikryverling:
@adamvoncorswant Regarding the licensing, JUG is using [Apache License 2.0](http://www.apache.org/licenses/LICENSE-2.0.html) and as far as I see it, using and modifying the source code as we're doing it here should be safe as long as you making a note of it (which I'm am in the JavaDoc). As we're using the same license for the toolbox I'm not sure if we need to provide a seperate copy just for the JUG code or if it's fine as is. 

@mikaellothman How did you reason about this when you we're implementing `DeterministicTimeUUIDFactory.createDateBits()` ? 

Licensing has never been my strongest side...

izettle-java/src/main/java/com/izettle/java/uuid/UUIDFactory.java
adamvoncorswant: I just want to reply on your first paragraph @erikryverling : 
Opinions differ :)  I believe that most java developers use package names too excessively instead of using proper naming on the actual class they are building. Packages are not solving much if you're not using them together with access modifiers to restrict outside access. Just using them for scoping is often a sign of otherwise bad design in my mind.

izettle-java/src/main/java/com/izettle/java/uuid/uuid1/UUID1Generator.java
mikaellothman: If you see it from a user of the library it has to be simple find when you look for a function for creating/generating a UUID. It's a bit hard to know which class to use given those names if you don't know the UUID-spec. What do you think about just one public UUIDFactory that supports all functions needed?

mikaellothman:
@erikryverling: To be honest, I didn't give the licensing enough thought when using the snippet.

adamvoncorswant:
@erikryverling : I feel that this PR is growing roots. Can we speed it up somehow? What about @mikaellothman 's last comments? Do you want to counter argument or amend according to the comments? 

A PR like this shouldn't have to wait for at most a couple of days before merged, given that all parties reply swiftly...

erikryverling:
I've digged into the licensing part a bit more, and I'm fairly confident that we have covered the things required from us.

adamvoncorswant:
Sounds good @erikryverling !

erikryverling:
@mikaellothman 
I've now refactored the package structure a bit to make it more concise and backwards compatible. Please have a look.

mikaellothman:
Awesome @erikryverling! Let's merge..

erikryverling:
Woho!

PR #102, opened 2015-01-27
Introduce checkstyle

adamvoncorswant:
ping @nzroller : sorry to bother you with this.

I'm not too sure about the pom configuration here. Should we depend on parent pom? I have not specified it here, but it seems it's picked up anyway from within the checkstyle tag config.

please advice.


nzroller:
I've inherited all from the base pom, since it's a public repo too I think it's ok, not great but ok.

Also I just set `<plugins><plugin>...checkstyle...` in the parent pom so it'll run in all projects by default.

nzroller:
You can see that the sub-poms don't need changing and the parent pom inherits plugin and repo configuration like most of our private repos. https://github.com/iZettle/izettle-toolbox/pull/102/files?w=1

adamvoncorswant:
Ah, didn't think about that the base pom is public. Awesome!

adamvoncorswant:
thanks @nzroller ! Slowly getting there...

PR #105, opened 2015-02-23
TLV work

fiddeandersson:
Fixed big blooper with multitag parsing.
Stricter parsing, more exception prone on malformed data.
Bunch of tests for multitag and expander tags.

Ping @staffanjonsson for another eye on this.


fiddeandersson:
Updated Staffan, thanks for catching my bugs...

PR #106, opened 2015-02-26
UUIDFactory: Fixes create alternate

adamvoncorswant:
Previously the method `createAlternative` mangled each byte in the UUID
without taking the structure of the data into consideration. The effect
of this was that the resulting alternative was not always a proper UUID
according to the standard.

This change does mainly two things:
* Make sure that `version` and `variant` is always preserved when
  creating an alternative.
* For time-based UUID, the entire MSB is preserved as we want the
  alternative to have the same time information as the original: only
  the LSB is altered

ping @erikryverling , as you've been involved in the UUID stuff lately.

related to https://github.com/iZettle/izettle-planet/pull/321

### DIFF
--- a/izettle-java/src/main/java/com/izettle/java/UUIDFactory.java
+++ b/izettle-java/src/main/java/com/izettle/java/UUIDFactory.java
@@ -1,5 +1,6 @@
 package com.izettle.java;
 
+import java.util.Arrays;
 import java.util.UUID;
 
 /**
@@ -7,6 +8,11 @@ import java.util.UUID;
  * @see <a href="https://tools.ietf.org/html/rfc4122">https://tools.ietf.org/html/rfc4122</a>
  */
 public final class UUIDFactory {
+
+    static final int VERSION_TIME_BASED = 1;
+    static final int VERSION_DCI_SECURITY = 2;
+    static final int VERSION_NAME_BASED = 3;
+    static final int VERSION_RANDOM = 4;
 
     private UUIDFactory() {
     }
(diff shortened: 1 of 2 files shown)


izettle-java/src/main/java/com/izettle/java/UUIDFactory.java
```diff
@@ -53,11 +59,41 @@ public static String createAlternative(String uuid) {
...
+                byte[] msBytes = longToBytes(mask(originalUuid.getMostSignificantBits(), mask));
+                //make sure to keep version and variant
+                msBytes[6] &= 0x0f;  //clear version
+                msBytes[6] |= originalVersion << 4;
+                msb = bytesToLong(msBytes);
+                break;
+            case VERSION_DCI_SECURITY:
+            case VERSION_NAME_BASED:
+            default:
+                throw new UnsupportedOperationException("Cannot create alternative for UUID with version: " + originalVersion);
```
adamvoncorswant: Note to reviewer: this might be a bit harsh, but I believe it's better to only handle the kinds of UUIDs that we actually are knowledgeable about. In the future we might want to consider supporting other types as well though.

xiaodong-izettle:
Just one note, if we want to release this, we'd better release this at the same time for both Sun and Purchase system. I have vague impression, that Refund feature depends on this method.

adamvoncorswant:
Thanks @xiaodong-izettle ! Let's investigate that after a possible merge of this. I've looked around quickly, but see no dependency expecting the alternate UUID to behave in any specific way.

xiaodong-izettle:
I mean the Refund feature somehow depend on that, when invoking calculating alternate UUID, both Sun and Purchase system SHOULD give the same result. :-)

As long as we upgrade library version at the same time, we are OK. :-)

adamvoncorswant:
Ok, I get it. Thanks @xiaodong-izettle !

izettle-java/src/main/java/com/izettle/java/UUIDFactory.java
erikryverling: Totally agree. It's  better to be safe than sorry.

adamvoncorswant:
Ok. I'm going to discuss IRL with @xiaodong-izettle about the logical consequences of this. BRR

adamvoncorswant:
Improved the javadoc as per discussion.

adamvoncorswant:
Ok. Turns out that we're in luck! The way the old version did mask the data in the `createAlternative`-method was by XOR:ing with 0x01. This accidentally meant that the version and the variant values was never touched and is still 4 and 2, respectively.

Hence: all UUIDs we have in our DB still have it's correct version. 

All good. @erikryverling : I'm ready to get this merged if you don't have any objections.

erikryverling:
Wow, I guess sometimes you need a little luck as :four_leaf_clover: 

Merging this!

adamvoncorswant:
Yeah, indeed @erikryverling :dancers:  Thanks for merging

PR #108, opened 2015-03-03
UUIDFactory: Fix for not preserving variant when createAlternative

adamvoncorswant:
Discovered when creating a test for backwards compatibility

Nice catch @mikaellothman ! This is all yours :) 

needs to be released and included in our other systems pronto before their release.


mikaellothman:
Thanks @adamvoncorswant for spending some extra time on this :)

PR #109, opened 2015-04-01
Add zero-padding parsing support in TLVDecoder.

fiddeandersson:
Adds a strict mode to the decoder, which defaults disabled. Note that in strict mode zero paddings with even sizes will produce valid TLV object (T=00 L=00 V:null). Which is OK I guess...?

Ping @staffanjonsson @oskarwiksten


oskarwiksten:
@fiddeandersson some checkstyle rule failed.

fiddeandersson:
Bah, thanks. I at least have format-on-save now...

oskarwiksten:
Good tests!

PR #110, opened 2015-04-06
JDBI Module, Binders currently duplicated in our services

freddd:
As we discussed before we should have a JDBI lib as these classes are generic and can be used by whoever, wherever (JDBI is used of course). These are basically copies of what's currently in accounting-service & translation-service. I had to make one small changes because I did not want to add guava as a dependency just to use Lists & Iterables. Also worth mentioning is that this module is overriding the default source & target, Java 7 and is instead using Java 8. It's a quick change if we would want this to be compiled using 7 instead.

The binders are used to automagic IN-clauses (*ArrayBinder) or change value to bind on (enums is a good example where the default is name() but sometimes we want ordinal()):
* ```EnumOrdinalBinder```
* ```EnumNameArrayBinder```
* ```IntegerArrayBinder```
* ```StringArrayBinder```

The ```PostgresArrayArgumentFactory``` is the ArgumentFactory that will intercept and do the actual binding.
The ```SqlArray``` captures the information to be bound.

Anyway, let me know what you think.

ping @aaanders @goncalosmo, @fredrikbackstrom, this might also be of interest for @adamvoncorswant as it goes into the toolbox which is OS.


izettle-jdbi/pom.xml
```diff
@@ -0,0 +1,41 @@
...
+    <properties>
+        <jdbi.version>2.59</jdbi.version>
+    </properties>
+
+    <dependencies>
+        <dependency>
+            <groupId>org.jdbi</groupId>
+            <artifactId>jdbi</artifactId>
+            <version>${jdbi.version}</version>
+            <scope>provided</scope>
```
adamvoncorswant: Does this mean that the dependency will be downloaded during build, but not included in the generated artifact?

izettle-jdbi/src/main/java/com/izettle/jdbi/PostgresArrayArgumentFactory.java
```diff
@@ -0,0 +1,39 @@
...
+            statement.setArray(position, ary);
+        };
+    }
+
+    private String getTypeName(final SqlArray<?> value) {
+
+        Object[] elements = value.getElements();
+        if (elements != null && elements.length > 0) {
+            if (elements[0] instanceof Number) {
+                return "integer";
```
adamvoncorswant: Is this really good enough? I mean, there are many types of numbers, such as `short, int, float, double`. Won't jdbi need to make the distinction?

adamvoncorswant:
I like this, and think it's the perfect place to put stuff like this.

izettle-jdbi/pom.xml
freddd: Yes, exactly. It will not be packaged in the izettle-jdbi jar but expects the container/jdk to have the dependency already. I assume that if someone is using the izettle-jdbi they also have the jdbi-jar already in their container, though I guess one could argue that they might use a different version but then we cannot guarantee that our extensions work as intended.

izettle-jdbi/src/main/java/com/izettle/jdbi/PostgresArrayArgumentFactory.java
freddd: I agree, we should change this. Did a quick check in the accounting service, we're creating an SqlArray of Longs which we are then here returning as type "integer". I'll look into it tomorrow.

adamvoncorswant: Check, thanks

fredrikbackstrom:
Nice things! @AndreasMeisingseth @tbaylis we might have things in risk repo that could go in here or at least copy-pastad here for use by others.

freddd:
Renamed the IntegerArrayBinder and added more granular clauses as discussed above. I will look into creating test cases using a dao so that we test the proper usage.

tbaylis:
Cool, once this is merged, I can look through which changes we can make in our code, and move some of our tools here aswell.

izettle-jdbi/src/main/java/com/izettle/jdbi/PostgresArrayArgumentFactory.java
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    @SuppressWarnings({"unchecked", "rawtypes"})
+    public boolean accepts(Class<?> expectedType, Object value, StatementContext ctx) {
+        return value instanceof SqlArray && clazz.isAssignableFrom(((SqlArray) value).getType());
+    }
+
+    public Argument build(Class<?> expectedType, final SqlArray<?> value, StatementContext ctx) {
+        return (position, statement, ctx1) -> {
+            final Array ary = ctx.getConnection().createArrayOf(getTypeName(value), value.getElements());
```
adamvoncorswant: As this is OS, I think we should take a bit more pride in variable naming. What about `elements`?

izettle-jdbi/src/main/java/com/izettle/jdbi/EnumNameArrayBinder.java
```diff
@@ -0,0 +1,22 @@
...
+import org.skife.jdbi.v2.SQLStatement;
+import org.skife.jdbi.v2.sqlobject.Bind;
+import org.skife.jdbi.v2.sqlobject.Binder;
+
+@SuppressWarnings("rawtypes")
+public class EnumNameArrayBinder implements Binder<Bind, Collection<Enum>> {
+
+    @Override
+    public void bind(SQLStatement<?> q, Bind bind, Collection<Enum> arg) {
+        Set<String> names =
```
adamvoncorswant: `final`?

izettle-jdbi/src/main/java/com/izettle/jdbi/EnumNameArrayBinder.java
```diff
@@ -0,0 +1,22 @@
...
+            arg
+                .stream()
+                .map(Enum::name)
+                .collect(Collectors.toSet());
```
adamvoncorswant: Why not the above 5 lines on one line?

izettle-jdbi/src/main/java/com/izettle/jdbi/PostgresArrayArgumentFactory.java
```diff
@@ -0,0 +1,45 @@
...
+    public PostgresArrayArgumentFactory(Class<?> clazz) {
+        this.clazz = clazz;
+    }
+
+    @SuppressWarnings({"unchecked", "rawtypes"})
+    public boolean accepts(Class<?> expectedType, Object value, StatementContext ctx) {
+        return value instanceof SqlArray && clazz.isAssignableFrom(((SqlArray) value).getType());
+    }
+
+    public Argument build(Class<?> expectedType, final SqlArray<?> value, StatementContext ctx) {
```
adamvoncorswant: Love that the `value` is `final`, but why not make all arguments `final` in this entire PR?

izettle-jdbi/src/main/java/com/izettle/jdbi/SqlArray.java
```diff
@@ -0,0 +1,36 @@
...
+    public static <T> SqlArray<T> arrayOf(Class<T> type, Iterable<T> elements) {
+        return new SqlArray<>(type, getListFromIterable(elements));
+    }
+
+    private static <T> List<T> getListFromIterable(Iterable<T> elements) {
+        List<T> listFromIterable = new ArrayList<>();
+
+        elements.forEach(listFromIterable::add);
+
+        return listFromIterable;
```
adamvoncorswant: Just:
```java
return StreamSupport.stream(iterable.spliterator(), false).collect(Collectors.toList());
```
?

izettle-jdbi/src/main/java/com/izettle/jdbi/SqlArray.java
```diff
@@ -0,0 +1,36 @@
...
+    public SqlArray(Class<T> type, Collection<T> elements) {
+        this.elements = elements.toArray();
+        this.type = type;
+    }
+
+    public static <T> SqlArray<T> arrayOf(Class<T> type, Iterable<T> elements) {
+        return new SqlArray<>(type, getListFromIterable(elements));
+    }
+
+    private static <T> List<T> getListFromIterable(Iterable<T> elements) {
```
adamvoncorswant: Naming: Just `toList`?

izettle-jdbi/src/test/java/com/izettle/jdbi/PostgresArrayArgumentSpec.java
```diff
@@ -0,0 +1,44 @@
...
+    @Test
+    public void testGetTypeName() {
+        assertThat(LONG_FACTORY.getTypeName(new SqlArray<>(Long.class, getList(1L)))).isEqualTo("bigint");
+        assertThat(LONG_FACTORY.getTypeName(new SqlArray<>(Integer.class, getList(1)))).isEqualTo("integer");
+        assertThat(LONG_FACTORY.getTypeName(new SqlArray<>(Double.class, getList(1.1)))).isEqualTo("float8");
+        assertThat(LONG_FACTORY.getTypeName(new SqlArray<>(Float.class, getList(1.1F)))).isEqualTo("float4");
+        assertThat(LONG_FACTORY.getTypeName(new SqlArray<>(String.class, getList("test")))).isEqualTo("text");
+    }
+
+    private <T> List<T> getList(T value) {
```
adamvoncorswant: Nitpicking naming here: this method name sounds like it's calling some unknown internal function on the `value` argument. Why not just name it `asList` instead?

izettle-jdbi/src/test/java/com/izettle/jdbi/PostgresArrayArgumentSpec.java
```diff
@@ -0,0 +1,44 @@
...
+        List<T> list = new ArrayList<>();
+        list.add(value);
+
+        return list;
```
adamvoncorswant: Why not just:
```java
return Stream.of(value).collect(Collectors.toList());
```

izettle-jdbi/src/main/java/com/izettle/jdbi/EnumNameArrayBinder.java
freddd: As I said in other PR's we need to take a look at the formatting for Java 8. It seems our formatter is doing crazy stuff sometimes.

goncalosmo: This one is my fault, though.

izettle-jdbi/src/main/java/com/izettle/jdbi/PostgresArrayArgumentFactory.java
freddd: Yes. I renamed it to ```elements```.

freddd:
Added .DS_Store to the .gitignore and also added individual versioning on izettle-jdbi, I don't really see a reason as to why it would have the same as the toolbox as a whole. Anyway, please look through it again and see if it's ready to be merged.

izettle-jdbi/pom.xml
```diff
@@ -0,0 +1,42 @@
...
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <parent>
+        <artifactId>izettle-toolbox</artifactId>
+        <groupId>com.izettle.toolbox</groupId>
+        <version>1.0.41-SNAPSHOT</version>
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+
+    <artifactId>izettle-jdbi</artifactId>
+    <version>1.0.0-SNAPSHOT</version>
```
nzroller: Version should stay the same as the toolbox proper I think (i.e. remove this line). For now it makes releases a tad easier.

izettle-jdbi/pom.xml
nzroller: Actually I think you'll run into issue with using `provided` here. The generally usage of this is with dropwizard apps and we shade all dependencies into the one uber jar. So there is no container and the jdk doesn't have it. I don't think it is the usual usage of this scope.

izettle-jdbi/src/main/java/com/izettle/jdbi/EnumNameArrayBinder.java
```diff
@@ -0,0 +1,18 @@
...
+
+@SuppressWarnings("rawtypes")
+public class EnumNameArrayBinder implements Binder<Bind, Collection<Enum>> {
+
+    @Override
+    public void bind(final SQLStatement<?> q, final Bind bind, final Collection<Enum> arg) {
+        final Set<String> names = arg.stream().map(Enum::name).collect(Collectors.toSet());
+        q.bind(bind.value(), SqlArray.arrayOf(String.class, names));
+    }
+}
```
nzroller: So this is truly minor but github is telling me something is wrong with not having a newline at the end of the file! I think git complains about that too... If you ever wondered why, the history is here: http://stackoverflow.com/questions/729692/why-should-files-end-with-a-newline and the first three answers all seem reasonable.

nzroller:
:+1:

izettle-jdbi/pom.xml
freddd: Alright, I'll change it. It's nice to not having it dependent on the toolbox though, as every time we make a change in a submodule we need to bump and release the toolbox even though it's quite rarely (if ever?) used as it's own entity as opposed to the submodules being used individually.

freddd:
@nzroller I just pushed a couple of commits that removes the individual versioning and adds new lines to two that were missing it. Please have a look again.

izettle-jdbi/src/test/java/com/izettle/jdbi/PostgresArrayArgumentSpec.java
```diff
@@ -0,0 +1,43 @@
...
+    public void testGetTypeName() {
+        assertThat(LONG_FACTORY.getTypeName(new SqlArray<>(Long.class, toList(1L)))).isEqualTo("bigint");
+        assertThat(LONG_FACTORY.getTypeName(new SqlArray<>(Integer.class, toList(1)))).isEqualTo("integer");
+        assertThat(LONG_FACTORY.getTypeName(new SqlArray<>(Double.class, toList(1.1)))).isEqualTo("float8");
+        assertThat(LONG_FACTORY.getTypeName(new SqlArray<>(Float.class, toList(1.1F)))).isEqualTo("float4");
+        assertThat(LONG_FACTORY.getTypeName(new SqlArray<>(String.class, toList("test")))).isEqualTo("text");
+    }
+
+    private <T> List<T> toList(T value) {
+        return Stream.of(value).collect(Collectors.toList());
```
xiaodong-izettle: Just use `Arrays.asList()`?

freddd:
Ready for merge?

nzroller:
Might want to do one last rebase here. Also `provided` scope is still there which I think will cause problems when we don't explicitly specify the jdbi dependency (for example when we get it as a transitive dependency from DW). I'm testing it to see though...

nzroller:
(woops I didn't rebase myself)...

freddd:
Alright, let me know regarding the provided scope. I generally think it's quite nice to have it as it makes our libs quite small but perhaps as you say, it will cause issues.

nzroller:
I double-checked and it actually didn't cause an issue. But I don't think we can make the assumption that it will be provided. In dropwizard apps no jar's are provided by the JDK and it's all shaded into one jar. If there is a situation where you need it provided I think it's best to then define the dependency and scope in that particular project.

PR #111, opened 2015-04-07
messaging: Extract body of QueueProcessingThread into QueueProcessingRunnable

oskarwiksten:
* Extracts the body of the `run()` method from `QueueProcessingThread` into a new class `QueueProcessingRunnable` that only implements `Runnable` (and not `Thread`).
* This would be useful when running a queue poller in an environment where the caller would not want to create the actual thread manually, such as in a java EE environment.


oskarwiksten:
Ping @xiaodong-izettle @mikaellothman

mikaellothman:
Nice refactoring @oskarwiksten! Check-styles doesn't agree though :(

oskarwiksten:
Thanks @mikaellothman . Fixed now.

mikaellothman:
Super!

PR #112, opened 2015-04-07
Upgrades AWS to v1.9.30 from v1.6.4

nzroller:
For awhile now Amazon have split their SDK into separate jar's per
service, so we are picking only the services we use here (SNS, SQS and
EC2).

For your consideration:
@oskarwiksten, @mxns, @xiaodong-izettle and @mikaellothman

### DIFF
--- a/izettle-cassandra/pom.xml
+++ b/izettle-cassandra/pom.xml
@@ -13,7 +13,7 @@
     <dependencies>
         <dependency>
             <groupId>com.amazonaws</groupId>
-            <artifactId>aws-java-sdk</artifactId>
+            <artifactId>aws-java-sdk-ec2</artifactId>
             <version>${aws.version}</version>
         </dependency>
         <dependency>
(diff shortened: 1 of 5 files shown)


nzroller:
FYI let's wait before merging this until dependant PR's are up because we'll get into trouble if we have different AWS versions packed in the final artifacts (since the `artifactId` has changed we may both v1.6.4 and v1.9.28.1 versions of classes)

xiaodong-izettle:
May I merge this one, @nzroller? :-)

xiaodong-izettle:
Actually, the newest version is `1.9.30` right now. :-)

nzroller:
ok @xiaodong-izettle I've changed the title and comments

xiaodong-izettle:
Thanks, @nzroller. I added some other stuff, could you please have a look? :-)

nzroller:
Looks good @xiaodong-izettle

xiaodong-izettle:
Thanks, @nzroller, merging...

PR #113, opened 2015-04-07
Add methods for setting up SQS subscription of a SNS topic

oskarwiksten:
* Adds class `AmazonSNSSubscriptionSetup` with a single public method `subscribeSQSQueueToSNSTopic` that sets up a SQS queue so that it receives all messages posted to a SNS topic.
* Requires the topic and queue to exist before creating the subscription.
* Sets up required permissions on the queue so that it can receive messages from the topic.

### DIFF
--- a/izettle-messaging/src/main/java/com/izettle/messaging/AmazonSNSSubscriptionSetup.java
+++ b/izettle-messaging/src/main/java/com/izettle/messaging/AmazonSNSSubscriptionSetup.java
@@ -1,0 +1,126 @@
+package com.izettle.messaging;
+
+import com.amazonaws.auth.policy.Policy;
+import com.amazonaws.auth.policy.Principal;
+import com.amazonaws.auth.policy.Resource;
+import com.amazonaws.auth.policy.Statement;
+import com.amazonaws.auth.policy.actions.SQSActions;
+import com.amazonaws.auth.policy.conditions.ConditionFactory;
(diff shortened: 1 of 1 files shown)


oskarwiksten:
Ping @nzroller @mikaellothman @xiaodong-izettle

izettle-messaging/src/main/java/com/izettle/messaging/AmazonSNSSubscriptionSetup.java
```diff
@@ -0,0 +1,126 @@
+package com.izettle.messaging;
+
+import com.amazonaws.auth.policy.Policy;
+import com.amazonaws.auth.policy.Principal;
+import com.amazonaws.auth.policy.Resource;
+import com.amazonaws.auth.policy.Statement;
+import com.amazonaws.auth.policy.actions.SQSActions;
+import com.amazonaws.auth.policy.conditions.ConditionFactory;
+import com.amazonaws.services.sns.AmazonSNSAsync;
```
nzroller: Doesn't have to be `Async` does it? I suppose we can use AmazonSNS, AmazonSQS interfaces instead?

nzroller:
nice @oskarwiksten !

mikaellothman:
Beatiful @oskarwiksten!

xiaodong-izettle:
Great! :+1:

PR #116, opened 2015-05-18
Find system TimeZoneId

goncalosmo:
ping @aaanders @adamvoncorswant 

I wonder if we have this logic somewhere else already?

I'm adding this because in prod we use `UTC` and in DEV we use `Europe/Stockholm`. 
When we use hibernate to fetch a `java.util.Date` it automatically parses it with the system's timezone, so we need to make sure that all the logic performed over that date uses the same timezone.

This is timezones though, so my head feels like a spaghetti carbonara by now. I might be overkilling it.


adamvoncorswant:
Just curious: why would you need the system time zone?

adamvoncorswant:
So, you don't need this anymore @goncalosmo , right?

goncalosmo:
Nope. As I suspected, I was overkilling it.
Thanks

PR #120, opened 2015-06-10
Uses Amazon{SNS,SQS} in subscription setup

nzroller:
- Use interfaces Amazon{SNS,SQS} rather than concrete asynchronous
  implementations for subscription setup

  We more likely want a synchronous subscription setup rather than
  asynchronous so this change gives us the possibility of passing in a
  regular client.

Ping @xiaodong-izettle


xiaodong-izettle:
Nice one, @nzroller. :+1:

PR #122, opened 2015-06-11
Expose more uuid utility functions

niklas-izettle:
* Create uuids as java.util.UUID

### DIFF
--- a/izettle-java/src/main/java/com/izettle/java/UUIDFactory.java
+++ b/izettle-java/src/main/java/com/izettle/java/UUIDFactory.java
@@ -37,6 +37,15 @@ public final class UUIDFactory {
     }
 
     /**
+     * Creates an UUID version 1. The UUID does NOT contain any host information, but only time information.
+     * The resulting string is guaranteed to be sufficiently unique to be usable as an id.
+     * @return The UUID.
+     */
+    public static UUID createUUID1() {
+        return UUID1Generator.generate();
+    }
+
+    /**
      * Creates a mutated URL safe (including removing trailing '=' characters) Base64 encoded UUID. The returned UUID
      * is based on the supplied one. This method is guaranteed to return a different UUID than the one supplied,
      * but done in a deterministic fashion (same value will be returned every time for each input value) and with the
@@ -79,7 +88,7 @@ public final class UUIDFactory {
             case VERSION_RANDOM:
                 byte[] msBytes = longToBytes(mask(originalUuid.getMostSignificantBits(), mask));
                 //make sure to keep version and variant
-                msBytes[6] &= 0x0f;  //clear version
+                msBytes[6] &= 0x0f; //clear version
                 msBytes[6] |= originalVersion << 4;
                 msb = bytesToLong(msBytes);
                 break;
@@ -90,7 +99,7 @@ public final class UUIDFactory {
         }
         //We can always change the least significant bits
         final byte[] lsBytes = longToBytes(mask(originalUuid.getLeastSignificantBits(), mask));
-        lsBytes[0] &= 0x3f;  //clear variant
+        lsBytes[0] &= 0x3f; //clear variant
         lsBytes[0] |= originalVariant << 6;
         final long lsb = bytesToLong(lsBytes);
         return toBase64String(new UUID(msb, lsb));
@@ -131,8 +140,8 @@ public final class UUIDFactory {
      * @param uuid The UUID to be converted.
      * @return The encoded UUID as string.
      */
-    static String toBase64String(UUID uuid) {
-        return toBase64String(asByteArray(uuid));
+    public static String toBase64String(UUID uuid) {
+        return toBase64String(toByteArray(uuid));
     }
 
     private static String toBase64String(byte[] bytes) {
@@ -140,7 +149,10 @@ public final class UUIDFactory {
         return result.split("=")[0]; // Remove trailing "=="
     }
 
-    static byte[] asByteArray(UUID uuid) {
+    public static byte[] toByteArray(UUID uuid) {
+        if (uuid == null) {
+            throw new IllegalArgumentException("uuid cannot be null");
+        }
         long msb = uuid.getMostSignificantBits();
         long lsb = uuid.getLeastSignificantBits();
         byte[] buffer = new byte[16];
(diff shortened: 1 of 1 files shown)


izettle-java/src/main/java/com/izettle/java/UUIDFactory.java
```diff
@@ -140,7 +158,10 @@ private static String toBase64String(byte[] bytes) {
         return result.split("=")[0]; // Remove trailing "=="
     }
 
-    static byte[] asByteArray(UUID uuid) {
+    public static byte[] asByteArray(UUID uuid) {
+        if (uuid == null) {
+            return null;
```
adamvoncorswant: Do you really think it's good to relax the stringency here? My take is that it's better to leave it to the client to do the null-checking. I wouldn't have said that a year ago, but since I've gotten used to `Optional` in java8, I almost think it's wrong for a method to handle nulls in this way...

izettle-java/src/main/java/com/izettle/java/UUIDFactory.java
```diff
@@ -37,6 +37,24 @@ public static String createUUID1AsString() {
     }
 
     /**
+     * Creates an UUID version 1. The UUID does NOT contain any host information, but only time information.
+     * The resulting string is guaranteed to be sufficiently unique to be usable as an id.
+     * @return The encoded UUID.
```
erikryverling: This is not encoded, so maybe just "The UUID"

izettle-java/src/main/java/com/izettle/java/UUIDFactory.java
```diff
@@ -149,16 +170,22 @@ private static String toBase64String(byte[] bytes) {
         return buffer;
     }
 
-    static UUID fromBase64String(String b64) {
-        if (b64 == null || b64.length() != 22) {
-            throw new IllegalArgumentException("Argument b64 string must be defined and have a length of exactly 22");
+    public static UUID fromBase64String(String b64) {
+        if (b64 == null) {
+            return null;
```
erikryverling: Varför retunerar vi `null` här`?

adamvoncorswant: Same here...

izettle-java/src/main/java/com/izettle/java/UUIDFactory.java
```diff
@@ -149,16 +170,22 @@ private static String toBase64String(byte[] bytes) {
...
         }
         return fromByteArray(Base64.b64StringToByteArr(b64));
     }
 
-    private static UUID fromByteArray(byte[] bytes) {
-        if (bytes == null || bytes.length != 16) {
-            throw new IllegalArgumentException("Argument byte array must be defined and have a length of exactly 16");
+    public static UUID fromByteArray(byte[] bytes) {
+        if (bytes == null) {
+            return null;
```
adamvoncorswant: And here...

izettle-java/src/main/java/com/izettle/java/UUIDFactory.java
niklas-izettle: Sure, I guess that makes sense!

izettle-java/src/main/java/com/izettle/java/UUIDFactory.java
erikryverling: Same as above

izettle-java/src/main/java/com/izettle/java/UUIDFactory.java
adamvoncorswant: @erikryverling : Hehe, seems we're commenting on the same things at the same time :)

izettle-java/src/main/java/com/izettle/java/UUIDFactory.java
```diff
@@ -140,7 +158,10 @@ private static String toBase64String(byte[] bytes) {
         return result.split("=")[0]; // Remove trailing "=="
     }
 
-    static byte[] asByteArray(UUID uuid) {
+    public static byte[] asByteArray(UUID uuid) {
```
adamvoncorswant: What about changing to `toByteArray` for consistency with the inverse method `fromByteArray`? I know it's a bit out of scope, but since the method wasn't public before no one outside was affected...

izettle-java/src/main/java/com/izettle/java/UUIDFactory.java
```diff
@@ -37,6 +37,24 @@ public static String createUUID1AsString() {
...
+    public static UUID createUUID1() {
+        return UUID1Generator.generate();
+    }
+
+    /**
+     * Creates an UUID version 1 as byte array. The UUID does NOT contain any host information, but only time information.
+     * The resulting string is guaranteed to be sufficiently unique to be usable as an id.
+     * @return The encoded UUID as a byte array.
+     */
+    public static byte[] createUUID1AsByteArray() {
```
adamvoncorswant: Do you think we need this, or should we simply be satisfied with the `fromByteArray` and the `toByteArray` methods instead? I have no strong opinion here, it's just that the calling code might look more symmetric with creating a `UUID` object, and then converting to and from `byte[]` using the two above methods instead. Perhaps. :)

Thoughts?

izettle-java/src/main/java/com/izettle/java/UUIDFactory.java
erikryverling: @adamvoncorswant 
Smurf! ;)

adamvoncorswant: Smurf back!

adamvoncorswant:
Other than the comments: really like this! This is how we should have done it from the beginning: just handling UUID objects instead of strings.

izettle-java/src/main/java/com/izettle/java/UUIDFactory.java
niklas-izettle: Sure, I think that would be cleaner.

niklas-izettle:
Updated according to review comments

adamvoncorswant:
:star:

adamvoncorswant:
I would have merged, but @erikryverling is assigned, so I'll leave it up to him.

erikryverling:
erik-approves

niklas-izettle:
haha, thanks!

PR #123, opened 2015-06-12
Set source level of java and cart modules to 1.6

erikryverling:
Nowadays the main source level of the toolbox is 1.8 (defined by it's parent pom).
However to prevent Android running into issues when using the toolbox we should set the source level of the used modules to 1.6

Ping @joezorry @silverpilen 

P.S Sorry for wall of types :wink:

### DIFF
--- a/izettle-cart/pom.xml
+++ b/izettle-cart/pom.xml
@@ -18,4 +18,15 @@
         </dependency>
     </dependencies>
 
+    <build>
+        <plugins>
+            <plugin>
+                <artifactId>maven-compiler-plugin</artifactId>
+                <configuration>
+                    <source>1.6</source>
+                    <target>1.6</target>
+                </configuration>
+            </plugin>
+        </plugins>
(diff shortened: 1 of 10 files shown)


joezorry:
Not to be a pain, but actually it's 1.7, maybe should have said that. But 1.6 works fine too :smile_cat:, it's not like it's the end of the world not using the diamond.

adamvoncorswant:
Is that really true? What about try-with-resources for example?

erikryverling:
Yeah, I think we need 1.6 to guard against  try-with-resources, or are you with another opinion @joezorry?

joezorry:
@adamvoncorswant @erikryverling you are correct sirs.  :+1:

erikryverling:
OK, cool. So @joezorry whenever you have time, please review

joezorry:
Seems fine to me :)

erikryverling:
Great!

PR #124, opened 2015-06-15
UUID: moved `getDateFromUUID1` from test to UUIDFactory

adamvoncorswant:
To make it publicly available for users of the library. Otherwise
outside callers need to know the values of `CLOCK_OFFSET` and
`CLOCK_MULTIPLIER`, which now can be kept private.

ping @erikryverling


erikryverling:
Great!

PR #128, opened 2015-08-27
CurrencyId: Add 11 missing currencies

adamvoncorswant:
Although quite rare (and some not even used anymore), we still want to
be able to interpret all possible currencies.

The added ones are:
```
AFN
AZN
GHS
MZN
RSD
SDG
SSP
TMT
VEF
ZMW
ZWL
```
Note: To keep the alphabetical order, all lines have been modified, so
this diff is easiest looked at using `--word-diff`


adamvoncorswant:
ping @oskarwiksten

oskarwiksten:
Running the following command : 
```
$ curl 'http://www.currency-iso.org/dam/downloads/lists/list_one.xml' \
      | xmlstarlet sel -t -m '//CcyNtry' -v 'Ccy' -o ', // ' -v 'CcyNm' -nl \
      | sort \
      | uniq
```

and matching against this list also suggest to add the following currencies:
* `SSP` - South Sudanese pound
* `VEF` - Venezuelan bolívar

You could also argue that `BTC` should be added to this list - but that's for another PR.

adamvoncorswant:
Thanks @oskarwiksten . Added SSP and VEF also. Rebased and force pushed.

oskarwiksten:
Nice @adamvoncorswant .

PR #129, opened 2015-08-31
UUIDFactory: Add method for parsing both string versions

adamvoncorswant:
As we're moving away from using the base64 representation of a `UUID`,
we still want to allow the old format in API's and as input parameters.
Testing for the two different versions everywhere makes the code ugly
and repetitive.

This change simply adds a parse method that will accept both versions of
the serialization format.

ping @niklas-izettle

### DIFF
--- a/izettle-java/src/main/java/com/izettle/java/UUIDFactory.java
+++ b/izettle-java/src/main/java/com/izettle/java/UUIDFactory.java
@@ -1,5 +1,7 @@
 package com.izettle.java;
 
+import static com.izettle.java.ValueChecks.empty;
+
 import java.util.Arrays;
 import java.util.Date;
 import java.util.UUID;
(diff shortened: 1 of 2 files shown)


niklas-izettle:
Isn't it better to be explicit on which version that is used?
If we some day would like to get rid of the old base64 format, then i guess this will make it harder?
So instead of having the flexibility of using both formats, then you have specifically state what format we are expecting.

adamvoncorswant:
I agree in theory. In reality, we will often stumble upon situations where both formats need to be allowed for backwards compatibility reasons (for example old links in emails and DTO's sent to other systems). This turned the code very ugly in quite a few places, so this is my suggestion to make it a bit prettier.

Of course, in the long run, I'd like all conversions between a string and `UUID` to be of the forms:
```
UUID.fromString
UUID::toString
```

niklas-izettle:
Makes sense! But I don't think we should use this method in places where we know that we will not need the old format?

adamvoncorswant:
I agree! Those are non-existant as of yet, but I intend to send out an email to everyone with info about this, so that they don't start doing b64 stuff when designing new endpoints etc. Then they should only rely on the default UUID conversion methods (eg, not use UUIDFactory at all).

niklas-izettle:
Cool :+1:

adamvoncorswant:
Thanks @niklas-izettle

PR #130, opened 2015-09-14
Introduce ServiceCharge.

MrShish:
* A service charge can either be fixed amount or a percentage applied on the final value of cart.
* It can also belong to its own vatGroupi

### DIFF
--- a/izettle-cart/src/main/java/com/izettle/cart/Cart.java
+++ b/izettle-cart/src/main/java/com/izettle/cart/Cart.java
@@ -28,6 +28,7 @@ public class Cart<T extends Item<T, D>, D extends Discount<D>, K extends Discoun
      * Produces a new immutable cart object from two lists if Items and Discounts
      * @param items the list of items, must not be empty (as a cart without items makes no sense)
      * @param discounts the list of cart wide discounts, possibly null or empty
+     * @param serviceCharge The applied service charge, possibly null
      */
     public Cart(List<T> items, List<K> discounts, S serviceCharge) {
         if (items == null) {
@@ -54,7 +55,7 @@ public class Cart<T extends Item<T, D>, D extends Discount<D>, K extends Discoun
         //Copy all items and discounts, but negate the quantities:
         List<T> inverseItems;
         List<K> inverseDiscounts;
-        S inverseServiceCharge = null;
+        S inverseServiceCharge;
 
         if (empty(itemLines)) {
             inverseItems = Collections.emptyList();
@@ -73,7 +74,9 @@ public class Cart<T extends Item<T, D>, D extends Discount<D>, K extends Discoun
             }
         }
 
-        if (!empty(serviceChargeLine)) {
+        if (empty(serviceChargeLine)) {
+            inverseServiceCharge = null;
+        } else {
             inverseServiceCharge = serviceChargeLine.getServiceCharge().inverse();
         }
 
--- a/izettle-cart/src/main/java/com/izettle/cart/Cart.java
+++ b/izettle-cart/src/main/java/com/izettle/cart/Cart.java
@@ -10,12 +10,15 @@ import java.util.Collections;
 import java.util.List;
 import java.util.SortedMap;
 
-public class Cart<T extends Item<T, D>, D extends Discount<D>, K extends Discount<K>> implements Serializable{
+public class Cart<T extends Item<T, D>, D extends Discount<D>, K extends Discount<K>, S extends ServiceCharge<S>>
+    implements Serializable {
 
     private final List<ItemLine<T, D>> itemLines;
     private final List<DiscountLine<K>> discountLines;
+    private final ServiceChargeLine<S> serviceChargeLine;
     private final long grossValue;
     private final Long discountValue;
+    private final Long serviceChargeValue;
     private final Long cartWideDiscountValue;
     private final Long actualVat;
     private final Double actualDiscountPercentage;
@@ -26,7 +29,7 @@ public class Cart<T extends Item<T, D>, D extends Discount<D>, K extends Discoun
      * @param items the list of items, must not be empty (as a cart without items makes no sense)
      * @param discounts the list of cart wide discounts, possibly null or empty
      */
-    public Cart(List<T> items, List<K> discounts) {
+    public Cart(List<T> items, List<K> discounts, S serviceCharge) {
         if (items == null) {
             items = Collections.emptyList();
         }
@@ -36,8 +39,10 @@ public class Cart<T extends Item<T, D>, D extends Discount<D>, K extends Discoun
         this.cartWideDiscountValue = CartUtils.getTotalCartWideDiscountValue(discounts, grossValue);
         this.discountLines = CartUtils.buildDiscountLines(discounts, grossValue, cartWideDiscountValue);
         this.itemLines = CartUtils.buildItemLines(items, grossValue, cartWideDiscountValue);
+        this.serviceChargeLine = CartUtils.buildServiceChargeLine(grossValue, cartWideDiscountValue, serviceCharge);
+        this.serviceChargeValue = CartUtils.getServiceChargeValue(grossValue, cartWideDiscountValue, serviceCharge);
         this.grossVat = CartUtils.summarizeGrossVat(itemLines);
-        this.actualVat = CartUtils.summarizeEffectiveVat(itemLines);
+        this.actualVat = CartUtils.summarizeEffectiveVat(itemLines, serviceChargeLine);
     }
 
     /**
@@ -45,10 +50,12 @@ public class Cart<T extends Item<T, D>, D extends Discount<D>, K extends Discoun
      * for refunds
      * @return the inversed cart
      */
-    public Cart<T, D, K> inverse() {
+    public Cart<T, D, K, S> inverse() {
         //Copy all items and discounts, but negate the quantities:
         List<T> inverseItems;
         List<K> inverseDiscounts;
+        S inverseServiceCharge = null;
+
         if (empty(itemLines)) {
             inverseItems = Collections.emptyList();
         } else {
@@ -65,7 +72,12 @@ public class Cart<T extends Item<T, D>, D extends Discount<D>, K extends Discoun
                 inverseDiscounts.add(discountLine.getDiscount().inverse());
             }
         }
-        return new Cart<T, D, K>(inverseItems, inverseDiscounts);
+
+        if (!empty(serviceChargeLine)) {
+            inverseServiceCharge = serviceChargeLine.getServiceCharge().inverse();
+        }
+
+        return new Cart<T, D, K, S>(inverseItems, inverseDiscounts, inverseServiceCharge);
     }
 
     /**
@@ -73,7 +85,7 @@ public class Cart<T extends Item<T, D>, D extends Discount<D>, K extends Discoun
      * @return the actual value of the cart
      */
     public long getValue() {
-        return grossValue - coalesce(cartWideDiscountValue, 0L);
+        return grossValue - coalesce(cartWideDiscountValue, 0L) + coalesce(serviceChargeValue, 0L);
     }
 
     /**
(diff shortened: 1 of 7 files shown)


MrShish:
ping @adamvoncorswant @ssprang @niklas-izettle

adamvoncorswant:
Lovely! I'll dive into this with pleasure!

ssprang:
Ah, nice! I thought we we're going to have some kind of meeting with ios and/or android to cement this first?

adamvoncorswant:
Added a small commit with minimal style changes (we're on java6 here, so the generics unfortunately has to be explicit).

Will review more...

izettle-cart/src/test/java/com/izettle/cart/CartTest.java
```diff
@@ -244,6 +292,40 @@ public void itShouldGroupVatWithDiscount() {
         );
     }
 
+    @Test
+    public void itShouldGroupVatWithDiscountAndServiceCharge() {
```
adamvoncorswant: Is this method really testing what it says it is? As far as I can see, we're only testing that the total VAT amounts add up? With and without discounts...

adamvoncorswant: Another note/question: should we include service charge in the grouped VAT matrix? I mean, what if we mix the service charge VAT together with the bananas, would that be ok?

MrShish: Thats a valid question that I also thought about while doing this. Its the most natural approach no?
I mean, if the serviceCharge has an vat amount, shouldn't it it be in the Vat  group list? nothing indicates thats the vatGroup should be product specific list.

Maybe @ssprang  and @aliciarissler has some input:
Should the vat amount of the service charge be included in the vat amount per vat percentage list?

MrShish:
and thanks for the java6 fix. 
Should put my ide to java 6 level on this project, didn't get any warnings.. Do you know if its all the projects in toolbox that is java6 or just cart?

adamvoncorswant:
@MrShish : it's only izettle-java (all the other uses default from the parent pom)

izettle-cart/src/main/java/com/izettle/cart/ServiceCharge.java
```diff
@@ -0,0 +1,32 @@
...
+     * The percent VAT that is applied to this service charge, can be null for situations where VAT is not applicable
+     * @return the vat percentage
+     */
+    Float getVatPercentage();
+
+    /**
+     * The amount that this service charge affects is target with
+     * @return The amount, can be null if the percentage is not
+     */
+    Long getAmount();
```
adamvoncorswant: Do you think that a service charge will ever have a predefined amount instead of a percentage?

izettle-cart/src/main/java/com/izettle/cart/CartUtils.java
```diff
@@ -215,7 +219,8 @@ static Double getDiscountPercentage(long grossValue, Long discountValue) {
         while (remainingDiscountAmountToDistribute != 0) {
             boolean reclaiming = remainingDiscountAmountToDistribute < 0;
             //We've distributed too much. reclaiming one at a time from the discounts with lowest roundingLoss
-            BigDecimal oldRoundingLoss = reclaiming ? discountIdxByRoundingLoss.firstKey() : discountIdxByRoundingLoss.lastKey();
+            BigDecimal oldRoundingLoss =
+                reclaiming ? discountIdxByRoundingLoss.firstKey() : discountIdxByRoundingLoss.lastKey();
```
adamvoncorswant: nah :)

izettle-cart/src/main/java/com/izettle/cart/CartUtils.java
```diff
@@ -275,7 +280,11 @@ static Double getDiscountPercentage(long grossValue, Long discountValue) {
         return totalGrossVat;
     }
 
-    static <T extends Item<T, K>, K extends Discount<K>> Long summarizeEffectiveVat(final List<ItemLine<T, K>> itemLines) {
+    static <T extends Item<T, K>, K extends Discount<K>> Long summarizeEffectiveVat(
+        final List<ItemLine<T, K>>
+            itemLines,
```
adamvoncorswant: while I like the splitting up into multiple lines, I also think that this is exaggerating :)

izettle-cart/src/main/java/com/izettle/cart/CartUtils.java
```diff
@@ -344,10 +362,85 @@ static Long calculateVatFromGrossAmount(final Long amountIncVat, final Float vat
...
+        if (!empty(serviceCharge.getAmount())) {
+            serviceChargeValue = serviceCharge.getAmount();
+        }
+
+        if (!empty(serviceCharge.getPercentage())) {
+            serviceChargeValue = round(actualValue * serviceCharge.getPercentage() / 100);
+        }
+
+        if (!empty(serviceCharge.getVatPercentage())) {
+            serviceChargeVat = serviceChargeValue - round(
```
adamvoncorswant: These two blocks of statements modify the same variable in a not too obvious way. If we're keeping the percentage on service charge, why don't we validate that it doesn't have both percentage and amount set at the same time (and change this code to use that fact)? Alternatively actually use both values in the calculation.

izettle-cart/src/main/java/com/izettle/cart/CartUtils.java
```diff
@@ -344,10 +362,85 @@ static Long calculateVatFromGrossAmount(final Long amountIncVat, final Float vat
...
+                (serviceChargeValue * 100) / (100 + (double) serviceCharge.getVatPercentage())
+            );
+        }
+
+        if (0 == serviceChargeValue) {
+            return null;
```
adamvoncorswant: Is this really correct? I mean, when we handle VAT, we make a difference between `null` and `0`. Could one imaging a situation where we have a service charge with value 0?

izettle-cart/src/main/java/com/izettle/cart/CartUtils.java
```diff
@@ -344,10 +362,85 @@ static Long calculateVatFromGrossAmount(final Long amountIncVat, final Float vat
...
+        return new ServiceChargeLine<S>(serviceCharge, serviceChargeValue, serviceChargeVat);
+    }
+
+    static Long getServiceChargeValue(
+        long grossValue,
+        Long cartWideDiscountValue,
+        ServiceCharge serviceCharge
+    ) {
+        if (empty(serviceCharge)) {
+            return null;
```
freddd: What about returning an ```Optional``` instead?

adamvoncorswant: We'd loved to, but this is java6 *without external dependencies*.

adamvoncorswant: So, in this library, we have to make a clear distinction between 0 and null.

izettle-cart/src/test/java/com/izettle/cart/CartTest.java
aliciarissler: From Xero's accountant expert: - Should this VAT be separated from the ”regular” VAT on sales? (i. e. in our reports in our app and when we import into Xero)
Yes, the user should have the option to choose which Account Code and VAT rate services charges are booked to. This is important because it might be a different VAT rate to food sales, and they might also want to report on this income differently. Also if they distribute the service charge out to their staff they'll need it booked to a different account code than sales to accurate track and account for giving money out to the staff. All of this applies to tips taken via iZettle too.

Will be solved when we add "categories" that they can use as account codes. As long as you can choose an account code for the service charge. That must work, right?

izettle-cart/src/main/java/com/izettle/cart/ServiceCharge.java
ssprang: @MrShish asked me this and I'm not 100% certain but I think it can be a predefined amount at for example hotels. Then it can also go under the name "service fee" according to this http://www.investopedia.com/terms/s/service-charge.asp . But maybe we should add it only when we know we need it? I guess this can always be solved as a product otherwise?

adamvoncorswant: @ssprang : thanks.

Well, even if it technically can be solved with a product, the difference is that we want this to be automatically added above bottom line on every cart for the customer. In that sense it would be nice to have it included as is. I guess that there is no harm?

izettle-cart/src/test/java/com/izettle/cart/CartTest.java
adamvoncorswant: But @aliciarissler , I din't think this really is what @MrShish asked about. I mean, we understand now that service charge needs to be flagged as special (to be able to distinguish this revenue from other), but when it comes to VAT: Do we need to treat the VAT differently, or should the VAT of the service charge be included in the table of VAT amounts on the receipt?

Eg, either:

| VAT Percentage | Amount |
|----------------------|------------|
|25.0% | 100 kr |
|25.0% | 13 kr |
|12.0% | 57 kr |

**Plus 12.0% VAT on service charge amounting to 13 kr**

or:

| VAT Percentage | Amount |
|----------------------|------------|
|25.0% | 100 kr |
|25.0% | 13 kr |
|12.0% | 70 kr |

aliciarissler: Ah, sorry. We asked that too: - Should the VAT on the service charge be displayed on the customer’s receipt too? 
Yes. It's not too important for general customers, but if the customer is claiming the cost back (for example if it's a business lunch), then they will need a valid VAT receipt with the correct breakdown.

But it's not really clear if its enough with grouping them together as you did in example 2. I'll ask.

adamvoncorswant: Thanks @aliciarissler (btw: updated my last comment to make the VAT percentages make more sense).

izettle-cart/src/main/java/com/izettle/cart/CartUtils.java
MrShish: what is?

adamvoncorswant: Perhaps just one line is enough for one argument, like so:
```java
final List<ItemLine<T, K>> itemLines,
```

MrShish: haha, now i get it. true! damn, autoformat!

izettle-cart/src/main/java/com/izettle/cart/CartUtils.java
MrShish: i could validate it here, and throw a illegal argument exception, but it shouldn't be possible to create a serviceCharge with both of them.
I think you did the same when calculating discount values. Ex [getNonRoundedDiscountValue] ( https://github.com/iZettle/izettle-toolbox/pull/130/files#diff-56b79a46332ac9d796c84edeccbe92b2R42)

MrShish: you know what! why just have either percentage or amount? why not just allow to have both? 
50kr + 20% etc..
gonna change.

izettle-cart/src/main/java/com/izettle/cart/CartUtils.java
MrShish: nice catch! will change!

izettle-cart/src/main/java/com/izettle/cart/CartUtils.java
adamvoncorswant: Exactly. And that's actually how it works in code you referenced (`getNonRoundedDiscountValue`).
 :)

izettle-cart/src/test/java/com/izettle/cart/CartTest.java
adamvoncorswant: I've been thinking a bit about this, and I tend to agree with @MrShish : I believe the VAT of the service charge should be an integral part of all other VAT amounts. So I'd guess that option 2 is the way to go.

MrShish:
@adamvoncorswant , modified the calculations to take in to account for both amount and percentage at the same time.

adamvoncorswant:
@MrShish : sweet. I'm happy with this now. Anyone else wants anything changed?

izettle-cart/src/main/java/com/izettle/cart/Cart.java
```diff
@@ -25,8 +28,9 @@
      * Produces a new immutable cart object from two lists if Items and Discounts
      * @param items the list of items, must not be empty (as a cart without items makes no sense)
      * @param discounts the list of cart wide discounts, possibly null or empty
+     * @param serviceCharge The applied service charge, possibly null
      */
-    public Cart(List<T> items, List<K> discounts) {
+    public Cart(List<T> items, List<K> discounts, S serviceCharge) {
```
niklas-izettle: Only one service charge?

MrShish: Si senior!

MrShish:
@adamvoncorswant : time to merge!

adamvoncorswant:
:shipit:

PR #131, opened 2015-09-27
Cart: Misc Cleanup

adamvoncorswant:
* Add `final` wherever possible, and change the code to reduce reassignments
* Remove unecessary method

Just cosmetic changes, no logic is supposedly changed: all tests classes are
left completely untouched

ping @MrShish : this might a tad too religious, but I kinda like it this way :)

### DIFF
--- a/izettle-cart/src/main/java/com/izettle/cart/CartUtils.java
+++ b/izettle-cart/src/main/java/com/izettle/cart/CartUtils.java
@@ -1,6 +1,7 @@
 package com.izettle.cart;
 
 import static com.izettle.java.ValueChecks.allEmpty;
+import static com.izettle.java.ValueChecks.allNull;
 import static com.izettle.java.ValueChecks.coalesce;
 import static com.izettle.java.ValueChecks.empty;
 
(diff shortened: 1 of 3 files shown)


izettle-cart/src/main/java/com/izettle/cart/CartUtils.java
```diff
@@ -387,57 +373,36 @@ static Long calculateVatFromGrossAmount(final Long amountIncVat, final Float vat
...
+        } else {
+            serviceChargeVat
+                = coalesce(serviceChargeValue, 0L)
+                - round((serviceChargeValue * 100) / (100 + (double) serviceCharge.getVatPercentage()));
         }
 
         return new ServiceChargeLine<S>(serviceCharge, serviceChargeValue, serviceChargeVat);
     }
-
-    static Long getServiceChargeValue(
```
MrShish: I know you can get the same value by getting the serviceChargeLine.getValue()
But i think this method is nice and consistent with the other ones, getValue(), getDiscountValue() etc..

MrShish:
otherwise looks good! nice cleanup. **finals** ftw!

izettle-cart/src/main/java/com/izettle/cart/CartUtils.java
adamvoncorswant: Fair enough. My problem was that the code was duplicated in the method above as well. So, the other solution would be to reintroduce the method below, but let `buildServiceChargeLine` call that instead of doing the calculation one more time. Would that be better?

MrShish: sure! sounds good!

adamvoncorswant:
Ok. Updated according to comment. Rebased and force pushed,

MrShish:
:+1:

adamvoncorswant:
Thanks @MrShish

PR #135, opened 2015-10-21
PublisherService: Handle all type of exceptions when failing to publish

MrShish:


MrShish:
Current implementation doesn't take in to account AmazonClientExceptions, thus, will miss if there is connectivity issues to aws. 
This will handle all types of unforseen exceptions, so if something exceptional happens: leave the message in the que.

MrShish:
ping @xiaodong-izettle

MrShish:
fyi: @ssprang

xiaodong-izettle:
@MrShish, I paired programming with @nzroller for this PR: https://github.com/iZettle/sun/pull/7580.

It will solve the issue at hand. But I think this one is good anyway. :)

xiaodong-izettle:
Nice and merging...

PR #137, opened 2015-11-10
JsonSerializer: add InstantModule to serialize as Instants.

aaanders:
Instants will now be serialized as ISO8601/RFC3339 formatted string like
`yyyy-MM-dd'T'HH:mm:ss.SSSZ`.

@nzroller can you check, please?

### DIFF
--- a/izettle-messaging/src/test/java/com/izettle/messaging/serialization/InstantModuleTest.java
+++ b/izettle-messaging/src/test/java/com/izettle/messaging/serialization/InstantModuleTest.java
@@ -1,0 +1,35 @@
+package com.izettle.messaging.serialization;
+
+import static org.junit.Assert.*;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import java.time.Instant;
+import java.time.ZoneId;
+import java.time.ZoneOffset;
+import java.time.format.DateTimeFormatter;
+import java.time.LocalDateTime;
+import org.junit.Test;
+
+public class InstantModuleTest {
+
+    private ObjectMapper objectMapper = new ObjectMapper().registerModule(new InstantModule());
+
+    @Test
+    public void testInstantSerializationDeserialization() throws Exception {
+        String instantString = objectMapper.writeValueAsString(Instant.EPOCH);
+        assertEquals("\"1970-01-01T00:00:00.000+0000\"", instantString);
+        Instant actualEpoch = objectMapper.readValue(instantString, Instant.class);
+        assertEquals(Instant.EPOCH, actualEpoch);
+    }
+
+    @Test
+    public void testAdamsBirthdaySerialization() throws Exception {
+        DateTimeFormatter format = DateTimeFormatter.ofPattern("yyyy-MM-dd'T'HH:mm:ss.SSSZ").withZone(ZoneId.of("UTC"));
+        Instant ctoEpoch = LocalDateTime.parse("1974-05-25T13:33:33.337+0000", format).toInstant(ZoneOffset.UTC);
+        String instantString = objectMapper.writeValueAsString(ctoEpoch);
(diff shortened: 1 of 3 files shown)


adamvoncorswant:
love it @aaanders !

izettle-messaging/src/test/java/com/izettle/messaging/serialization/InstantModuleTest.java
```diff
@@ -0,0 +1,21 @@
...
+
+    @Test
+    public void testInstantSerializationDeserialization() throws Exception {
+        ObjectMapper objectMapper = new ObjectMapper();
+        objectMapper.registerModule(new InstantModule());
+
+        String instantString = objectMapper.writeValueAsString(Instant.EPOCH);
+        assertEquals("\"1970-01-01T00:00:00.000+0000\"", instantString);
+        Instant actualEpoch = objectMapper.readValue(instantString, Instant.class);
+        assertEquals(Instant.EPOCH, actualEpoch);
```
adamvoncorswant: You have to test if my birthday will be correctly converted as well '1974-05-25' Aight?

aaanders: works!

izettle-messaging/src/test/java/com/izettle/messaging/serialization/InstantModuleTest.java
```diff
@@ -0,0 +1,35 @@
...
+    }
+
+    @Test
+    public void testAdamsBirthdaySerialization() throws Exception {
+        DateTimeFormatter format = DateTimeFormatter.ofPattern("yyyy-MM-dd'T'HH:mm:ss.SSSZ").withZone(ZoneId.of("UTC"));
+        Instant ctoEpoch = LocalDateTime.parse("1974-05-25T13:33:33.337+0000", format).toInstant(ZoneOffset.UTC);
+        String instantString = objectMapper.writeValueAsString(ctoEpoch);
+        assertEquals("\"1974-05-25T13:33:33.337+0000\"", instantString);
```
adamvoncorswant: Lol. :cake:

aaanders:
So agreed with @nzroller that we should use https://github.com/FasterXML/jackson-datatype-jsr310 but for now there are dependency issues. I think @xiaodong-izettle will upgrade jackson in the wildfly9 branch.

adamvoncorswant:
I can't seem to find information that the referenced jackson update supports `Instant`?

aaanders:
It has a JavaTimeModule that is supposed to support the entire java.time package as I understand it. Will try it out after lunch.

aaanders:
Just tried out the JavaTimeModule and the result doesn't contain milliseconds which is a bit sad. Also the time zone info/offset is poor. 

Expected :"1970-01-01T00:00:00.000+0000"
Actual   :"1970-01-01T00:00:00Z"

adamvoncorswant:
Sad, but great that you found out! Let's use this PR then?! @nzroller ?

aaanders:
So, that wasn't the entire truth. Stupid test case. When the millis part is `.000` it won't be included at all. I guess it can be considered an optimisation. And the Z simply means no TZ and no offset which is the case for Instant.

So, here are some examples:
```
Instant epoch          > "1970-01-01T00:00:00Z"
Instant epochPlus555   > "1970-01-01T00:00:00.555Z"
ZonedDateTime epochZ   > "1970-01-01T00:00Z[UTC]"
ZonedDateTime epochCet > "1970-01-01T01:00+01:00[CET]"
```

With this in mind I think it might be worth it to  use the JavaTimeModule(or JSR-310 module until we bump jackson again). Then we would have everything set for serializing the `java.time` package and also not having to maintain this throughout our services. Thoughts?

adamvoncorswant:
Sounds like a plan!

PR #136, opened 2015-10-30
Metadata to publish artifacts to maven central

nzroller:
- Inherits plugin information from izettle-oss POM
- Fixes to pass new new-line-at-end-of-file checkstyle rule
- Adds Travis CI configuration

Tested locally by running:

```
docker run -it --rm --name my-maven-project -v "$PWD":/usr/src/mymaven -w /usr/src/mymaven maven:3 mvn clean install
```

PR #141, opened 2015-11-19
Checkstyle: Bump oss basepom to 1.9

adamvoncorswant:
To get the latest checkstyle rules:

* Comma(,) at end of line
* Plus(+) and dot(.) at beginning of line.
* All files should have an NL at end of file

ping @nzroller again: did I get it right with the oss pom?

### DIFF
--- a/izettle-cart/src/main/java/com/izettle/cart/CartUtils.java
+++ b/izettle-cart/src/main/java/com/izettle/cart/CartUtils.java
@@ -44,11 +44,12 @@ class CartUtils {
             retVal = BigDecimal.valueOf(discount.getAmount());
         }
         if (discount.getPercentage() != null) {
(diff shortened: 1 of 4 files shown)


nzroller:
Yes except see https://github.com/iZettle/izettle-toolbox/pull/136 -- basically I'm the only one who can release right now because I'm the only one with write permissions on sonatype for the groupId `com.izettle`. I was hoping I could automate the whole process somehow but haven't got around to it yet...

nzroller:
So for this PR, I think it's still fine to use the "private" basepom and I'll try to figure it out today...

nzroller:
I updated the commit back to `izettle:1.7` -- @sirtik is helping test and implement semi-automated OSS releases and once that is ready we can make the switch to the oss pom via #136

PR #144, opened 2015-12-02
Add utility methods to exact timestamp information from UUID1.

xiaodong-izettle:
Ping @adamvoncorswant. Thanks.


adamvoncorswant:
But we only support java6 in the toolbox?

izettle-java/src/main/java/com/izettle/java/UUIDFactory.java
```diff
@@ -269,4 +271,36 @@ public static UUID generate() {
...
+
+    /**
+     * Return the unix timestamp contained by the provided time-based UUID.
+     *
+     * @param uuid the UUID to return the timestamp of.
+     * @return the unix timestamp of {@code uuid}.
+     *
+     * @throws IllegalArgumentException if {@code uuid} is not a version 1 UUID.
+     */
+    public static long unixTimestamp(UUID uuid) {
```
adamvoncorswant: Also, if we're only going to use this for `Instant` should we not actually use the nanosecond units as well?

xiaodong-izettle:
Well, as TC said OK, I assume we support Java-8 already in toolbox. :-)

adamvoncorswant:
Interesting. The pom for the izettle-java project specifies 1.6...

adamvoncorswant:
I'm up for discussing this though. Only reason we have java6 is because of Android. Most of the code they could probably copy to their repo, but especially the cart code seems a bit risky to maintain two version of. Thoughts?

xiaodong-izettle:
Sure, fixed. Removed Java-8 stuff, replaced with Java-6 equivalent.

PR #145, opened 2015-12-18
Added Class Assert to Java module. 

pgisslen:
Assert have similar functionality as Guava Precondition but with some extra features.

Added corresponding JUnit test as well.


pgisslen:
@adamvoncorswant @mikaellothman Ping

izettle-java/src/main/java/com/izettle/java/Assert.java
```diff
@@ -0,0 +1,200 @@
...
+    }
+
+    /**
+     * Assert that an object is {@code null} .
+     * <pre class="code">Assert.isNull(value, "The value must be null");</pre>
+     * @param object the object to check
+     * @param message the exception message to use if the assertion fails
+     * @throws IllegalArgumentException if the object is not {@code null}
+     */
+    public static void isNull(Object object, String message) {
```
adamvoncorswant: Hehe. I'm trying to come up with a scenario when you would want to assure this...

adamvoncorswant:
Interesting. It's like an assert version of `ValueChecks`. Would it be possible to reduce the amount of code by calling `ValueChecks` more inside the assert methods?

also pinging @fredrikbackstrom because you probably have an opinion here.

izettle-java/src/main/java/com/izettle/java/Assert.java
pgisslen:
```
public void saveId( long id ) {
   Assert.isTrue( id == 42, "Id must have length 42);
   //Do stuff
}

**Comment by pgisslen** [comment]
I tried to use as much as possible from ValueChecks, did you have anything particular i mind that I missed?

**Comment by fredrikbackstrom** [comment]
Good stuff @pgisslen, here are my thoughts! 

I never got into using value checks because I've always felt that we should never need to check for null but use optionals (guava prior to java 8). The other features of ValueChecks like empty and such I feel we should not use cause they do to much and are rarely needed. 

I'm also  a fan of Guavas Preconditions and use them extensively for tight checks in constructors to assert nothing is ever null. I know there is some debate over Guavas be or not to be, maybe people think it's a new Apache Commons. For our Dropwizard Services I feel we should stick to Guava as it is bundled and not ValueChecks/Assertions. In most cases you can also get away from any library by using Objects.requireNonNull. This and using Optional from Java 8 may render both libraries obsolete, depends on how much features you want from the preconditions.

For those that like ValueChecks and the SUN project these Assertions are a nice addition!

I'm not so sure about having these many specific assertions though, when adding specific methods for all cases, classes like these tend to grow a bit too big.

My view is that all we need is the isNotNull and isTrue. These two can solve all other cases much like Guavas checkNotNull and checkArgument. 

What's your thoughts on that?

File: izettle-java/src/main/java/com/izettle/java/Assert.java [hunk 2]
```diff
@@ -0,0 +1,200 @@
...
+
+    /**
+     * Assert that an array is not null and has no null elements.
+     * Note: Does not complain if the array is empty!
+     * <pre class="code">Assert.noNullElements(array, "The array must have non-null elements");</pre>
+     * @param array the array to check
+     * @param message the exception message to use if the assertion fails
+     * @throws IllegalArgumentException if the object array contains a {@code null} element
+     */
+    public static <T> T[] noNullElements(T[] array, String message) {
```
adamvoncorswant:
``` 
if (ValueChecks.anyNull(array)) {
    throw new IllegalArgumentException(message);
}
```

izettle-java/src/main/java/com/izettle/java/Assert.java
```diff
@@ -0,0 +1,200 @@
...
+        if (ValueChecks.empty(map)) {
+            throw new IllegalArgumentException(message);
+        }
+        return map;
+    }
+
+    /**
+     * Assert a boolean expression, throwing {@code IllegalStateException}
+     * if the test result is {@code false}. Call isTrue if you wish to
+     * throw IllegalArgumentException on an assertion failure.
```
adamvoncorswant: I don't get this comment. Copy paste error?

izettle-java/src/main/java/com/izettle/java/Assert.java
```diff
@@ -0,0 +1,200 @@
+package com.izettle.java;
+
+/*
+ * Copyright 2002-2015 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
```
adamvoncorswant: Are there other original authors than you @pgisslen ? If so, haven't we changed the code a tad too much? If we're the only authors, we should remove the licence section altogether (as we don't have them in the other toolbox source files).

pgisslen: Yes, should remove that. Most of the original is changed, it's mainly the doc that is still there.

pgisslen:
@fredrikbackstrom Let's have a discussion on beginning of next week if you're there. I basically agree with you, especially regarding the use of Optional. 

 The reason for I want to add this functionality is that not PreConditions and Objects have some good stuff but not everything. The approach with this Assert class is:
- Gather all checks we feel useful in one place, basically combine PreCondition/Objects and Valuechecks.
- Force user to always add a reason when Assert fails. I think "no reason" assertions is quite useless. The purpose is to add more context information, otherwise we can just wait for the to occur later in the code.

pgisslen:
Another option would be to add the assert in ValueChecks, basically additional methods (same name but prefixed with assert like assertNotEmpty assertNoneNull etc

adamvoncorswant:
I'm up for either suggestions. What's your preference @pgisslen ?

pgisslen:
After a IRL-discussion(!), we decided to go for the merged solution, i.e.. moving the functionality to ValueChecks class and prefix the method with assert.

pgisslen:
I have made the changes we discussed.

izettle-java/src/main/java/com/izettle/java/ValueChecks.java
```diff
@@ -148,4 +148,145 @@ public static boolean allEmpty(Object... objects) {
...
+
+    /**
+     * Assert that the given String has valid text content; that is, it must not
+     * be {@code null} and must contain at least one non-whitespace character.
+     * <pre class="code">Assert.hasText(name, "'name' must not be empty");</pre>
+     * @param text the String to check
+     * @param message the exception message to use if the assertion fails
+     * @throws IllegalArgumentException if the text does not contain valid text content
+     */
+    public static String assertNotBlank(String text, String message) {
```
adamvoncorswant: Just to keep the same syntax: should we call this `assertNotEmpty` instead?

adamvoncorswant: If you change the name, we could use the same method for `String`, `Collection` and `Map`. Much the same as the current `empty` method does.

izettle-java/src/main/java/com/izettle/java/ValueChecks.java
```diff
@@ -148,4 +148,145 @@ public static boolean allEmpty(Object... objects) {
...
+
+    /**
+     * Assert that an Collection is not null and has no null elements.
+     * Note: Does not complain if the Collection is empty!
+     * <pre class="code">Assert.assertNoNulls(array, "The Collection must have non-null elements");</pre>
+     * @param collection the Collection to check
+     * @param message the exception message to use if the assertion fails
+     * @throws IllegalArgumentException if the object Collection contains a {@code null} element
+     */
+    public static <T> Collection<T> assertNoNulls(Collection<T> collection, String message) {
```
adamvoncorswant: Should we add a corresponding non-assert method first, which this method then can call (as in most of the other assert-methods)? Eg. `anyNull(Collection<T> collection)`

adamvoncorswant:
Except for comments, I think this is really good.

mikaellothman:
First sorry for the late response,  it came some :ski:, :santa: and :christmas_tree: in between... 

Just to fill in on the background of `ValueChecks`, it was as @fredrikbackstrom mentioned introduced before Java 8 and Optionals. That probably implies that we by now should deprecate some of the methods in this class. 

I'm not a frequent user of Guava and I think it also have a bunch of things that should be deprecated by now.

So, to summarize I think this is a nice improvement and that it should be merged as soon as the comments are fixed :+1:

izettle-java/src/main/java/com/izettle/java/ValueChecks.java
```diff
@@ -148,4 +148,145 @@ public static boolean allEmpty(Object... objects) {
     public static <T, S extends T> T ifEmpty(T subject, S fallback) {
         return empty(subject) ? fallback : subject;
     }
+
+    /******************************************************************
```
mikaellothman: This is only my personal opinion, but I think this kind of blocks bloats the code and doesn't give you that much...

adamvoncorswant: I used to be on your side @pgisslen , but after spending some time with @mikaellothman , I'm now a convert :)

pgisslen: I didn't write the ifEmpty method, it already exist, should I remove it you mean?

adamvoncorswant: Nah, we're just discussing the trivial matter of the separator: 
```
/******************************************************'
```
 :)

mikaellothman:
Ok. I think this is good to merge. Nice work @pgisslen!

adamvoncorswant:
:+1:

PR #146, opened 2016-01-13
Local queue

fredrikbackstrom:
Implementation of a local Queue based on H2 DB. This allows
us to choose between an in-memory queue or a file based queue.
The H2 db is quite performant and should suffice for us to begin
with. Using the H2 database makes the FIFO queue implementation easy.
An improvement to speed things up is to implement a simple FIFO Q
as a RandomAccessFile, but this is an improvement that I don't
think we need at the moment.

I wanted the Local Queue to be injected into SUN code without too
much change needed, hence I have baked the Queue into the
MessagePublisher interface such that it can be used in SUN:s
MessagingManager class without too much change. This however made
the task a bit bigger, in fact most changes in this commit is not
connected to the Q itself but the surrounding classes.

If we fail to send a message via a MessagePublisher it can be due
to many reasons but the two big ones are Network issues towards SNS
and serialization issues for the messages being sent. If there is a
network problem then we know for sure that once the network is up we
will be able to send the message. If there is a serialization issue we
need to make a code release in order to fix that. In any case messages
that cannot be handled are removed from the main Q after X number of
retries and pushed onto a DLQ queue. The DLQ is just a normal Q that
is polled less frequently. This feature will make it possible to remove
events from the top of the Q that cannot be handled and risk blocking
events behind them from being processed. The DLQ has a higher number
of retries before being evicted in order to have a chance to handle
the issue before it is removed. 

The introduction of the Q also introduces the fact that we need
threads to poll both the Q and DLQ in order to push the locally stored
events onto the SNS:s again. As there is one Q per SNS topic we will get
quite some queues to pull, to not need one thread per topic a
ExecutorServices is used in order to control the number of threads
polling Q:s

The class TaskQueueMessagePublisherFactory is the factory that
creates a MessagePublisher backed with a Q and destination
MessagePublisher.

Not sure about having the Queue Implementation in this repo as it introduces
dependency on H2. Let's discuss.


adamvoncorswant:
Interesting stuff, nice work @fredrikbackstrom !

What's your thought of using a random access file instead of H2? That would remove the H2 dependency, possibly at the cost of messier code?

fredrikbackstrom:
I'm all for the RAF as the final solution, it would be very nice. This is a bit of a quick'n'dirty version as it is very easy to implement with sql. 

The messier byteshuffling code would only be in one class so I don't think that's an issue. It's more of taking the time to implement it and write a few more tests to be certain that we handle all file based issues that may occur.

xiaodong-izettle:
Interesting stuff, @fredrikbackstrom. 

Will this replace SQS?

fredrikbackstrom:
@xiaodong-izettle no this will not replace SQS:s. It's ment as a fallback when SNS:s do not respond. Messages will be written locally and then a thread will push them onto the SNS again. It's mainly to recover from SNS outages. The same feature can be added infront of the SQS publishers as well if needed.

izettle-messaging/src/main/java/com/izettle/messaging/queue/RandomAccessTaskQueue.java
```diff
@@ -0,0 +1,386 @@
+package com.izettle.messaging.queue;
```
fredrikbackstrom: ping @adamvoncorswant I could not not resist trying it out yesterday. The tricky part is handling deletes in the file but I think the relaxation of that problem as described in the JavaDoc will work. This is not final solution as a main method exists and no tests are written for it. I just want to have your eyes on it

xiaodong-izettle:
Thanks for the explanation, @fredrikbackstrom. 

If it's only for SNS outage, any reason we need a separate local queue? We could just persist those failed to be published messages either into a file, or a db table directly (could be a separate RDS just for message persisting), and have some background thread/process/service whatever to handle those messages, yes?

That should simplify the implementation a bit?

xiaodong-izettle:
By the way, thread safety would be crucial here, as we do assume `MessagePublisher` implementation to be thread safe...

fredrikbackstrom:
A RDS does not make it simpler I'd say. In order to use a RDS you just change the SQL in the statement manager but all the queueing and polling i still needed. A RDS is also a resource not on the actual instance and may also fail and that sort of breaks the idea of being more resillient against different types of AWS outages.

fredrikbackstrom:
And for thread safety this will ofcourse be thread safe for multiple writers. But polling from the Queue is not safe to do for multiple threads as they might read the same messages.

xiaodong-izettle:
Well, I would say RDS should at least be more resilient then a memory DB such as H2, yes? :)

And RDS could run in multiple availability zone, and I would say the chance that RDS and SNS fail at the same time should be really small, yes? 

You're right we should be more resilient against different type of AWS outages, but we do not need to plan for different AWS outages happen at the same time, at least not for now, yes? 

As regards for guarantees of queue polling, it depends on how we designing the consuming side...

xiaodong-izettle:
And H2 does have file persistence mode, I think we should go for that mode rather than memory mode, yes?

fredrikbackstrom:
If you want to be as resilient as possible you should first commit to something local compared to a resource outside of your control. For me this is quite a tiny thing to write to a local file and then consuming that. Setting up RDS:s for this feels like a quite big thing architecturally and operational. 

Both inMemory and file are implemented. inMemory is only used for tests, files are implemented for production. Still you get to choose. 

"As regards for guarantees of queue polling, it depends on how we designing the consuming side..."

of course it depends on how you design the consuming side. In this case the design is that one thread is more than enough to poll the queue and we also use one messagepublisher per message type in sun so there is no need to have multiple consumers of the queue, it's not what it is design for.

xiaodong-izettle:
Well, writing to a local file and consuming it is not easy at all... Too many things can go wrong... Actually, I would recommend go for some battle tested open source project for that, if we really want to use file persistence and want to play safe.

IMHO, local does not mean resilient, and it does not mean it's resilient than things out of our control... AWS is quite robust, and good SLA. And it is definitely more robust and resilient than a single local machine to me...

From my perspective, I would recommend we use H2 memory/file storage for local, and RDS for prod. We just insert messages into db when we could not publish them immediately, so that they could be handled later, to make the implementation simpler.

fredrikbackstrom:
The local queue is a fallback queue in order to survive outages that we hope are rare. If you think having a local file is too big of an change I would say we should have it as an inmemory queue for as long as the instance is running or nothing at all. If it is restarted then those messages are lost but at least we are safer than we are today. Adding messages to a RDS is not what I'm after here at all so then I'd rather skip this type of feature entirely and close the PR which is fine by me.

adamvoncorswant:
We've had discussion around this conceptual problem quite a few times, and time and again it's been brought up that it would be nice to have generic way of securing messages. The simplest, most resilient way of doing that is to write the messages locally as a backup before posting it to the message queue (SNS). Almost what we have in the planet today, but something that works out of the box for anyone using the client and doesn't require yet another database.

If SNS/SQS is down or unreachable, it can be for a number of reasons, network problems being one of them. Then you don't want to have your fallback over the network (RDS or similar). Local files or in memory is the only robust alternative here, as far as I can see.

adamvoncorswant:
That said, I think this is a good way to go. Will still have to look at the PR though :)

xiaodong-izettle:
Thanks for the comment, @adamvoncorswant. 

Why db is not an option? As far as I could see, the mechanism we have in Planet works quite well. It never loses any message, yes? Isn't that what we want? 

As for the network problem, except prod-app1, which runs schedule jobs stuff, we almost do not need to publish messages when the server is cut off from network, yes? And RDS could span several availability zones, and failure across several availability zone is really really rare... 

Another thing, why should we backup every message before we send it, rather than only persist it when we fail to deliver? If it's for event sourcing, using a file is a bit rudimentary, comparing with using a DB, yes?

adamvoncorswant:
Yes, it's working fine in the planets, but I think that's too heavyweight for being the default behaviour in a messaging client.
And, again, if you fail to talk to an external service, I think it's better to have the fallback locally instead of in another remote service.

Think about a desktop email client. If it fails to send the email to it's configured SMTP server, it doesn't try to send the email somewhere else for later retrieval and then resend, right? Rather, it keeps the email locally until the SMTP server is available again.

fredrikbackstrom:
@xiaodong-izettle just to clarify this is a fallback and not the default. If the SNS responds then it is never written to queue. This is for the times when there is an actual issue. A lightway feature to be used as a drop-in feature whenever you push to sns as a safe-guard. If like planets you have to do everything you can to make sure that msgs are never lost than you need to take it to the next level.

xiaodong-izettle:
Thanks for the comment and clarification, @fredrikbackstrom and @adamvoncorswant. 

I would still want to try that, how about we just use H2 db with file persistence, so that we raise the abstraction level a bit. Rather than directly talk to a file, we speak SQL... :)

ssprang:
I think @chas678 and @pgisslen would be happy to have a look at this as well.

pgisslen:
I'll pick this up now :)

fredrikbackstrom:
@pgisslen nice, we can have a chat about it if you want. 

I also like to mention that the issue is almost gone now that we have moved sun to AWS. We have promille of events not able to be sent on the queue so there are probably other things that could be a priority.

adamvoncorswant:
Sad...

fredrikbackstrom:
Yeah, sometimes very interesting things go to die. It was not in vain but I would not promote this today.

PR #149, opened 2016-01-27
Catch amazon service exception when sending batch to queue

karlgrund:
Methods in QueueServiceSender now catch AmazonServiceException which might be thrown if Amazon-service is offline or URL is faulty.
Cleaned up test case for QueueServiceSender where EmptyBatchRequestException is never thrown.
Renamed test to better represent what it does.


AndreasMeisingseth:
:+1:

mikaellothman:
Great @karlgrund, thanks for fixing!

PR #150, opened 2016-02-02
Dependency Mgmt: Remove izettle-cart and izettle-emv's dependency on izettle-java

adamvoncorswant:
As that project will start using java8 features, and these projects are
used by the android platform.

ping @softarn perhaps?

### DIFF
--- a/izettle-cart/src/main/java/com/izettle/cart/CartUtils.java
+++ b/izettle-cart/src/main/java/com/izettle/cart/CartUtils.java
@@ -1,10 +1,5 @@
 package com.izettle.cart;
 
-import static com.izettle.java.ValueChecks.allEmpty;
-import static com.izettle.java.ValueChecks.allNull;
-import static com.izettle.java.ValueChecks.coalesce;
-import static com.izettle.java.ValueChecks.empty;
-
 import java.math.BigDecimal;
 import java.math.RoundingMode;
 import java.util.*;
@@ -86,7 +81,7 @@ class CartUtils {
             }
         }
 
-        if (!allEmpty(lineItemDiscount, cartWideDiscount)) {
+        if (lineItemDiscount != null || cartWideDiscount != null) {
             return coalesce(cartWideDiscount, 0L) + coalesce(lineItemDiscount, 0L);
         }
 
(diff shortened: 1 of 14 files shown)


adamvoncorswant:
ping also @FredrikMunter , as izettle-emv is your baby

izettle-cart/src/main/java/com/izettle/cart/CartUtils.java
```diff
@@ -279,7 +274,7 @@ static Double getDiscountPercentage(long grossValue, Long discountValue) {
             }
         }
 
-        if (!empty(serviceChargeLine) && !empty(serviceChargeLine.getVat())) {
+        if (serviceChargeLine != null && serviceChargeLine.getVat() != null) {
```
FredrikMunter: Good to remove deps - I approve! I take it these cannot in any way be empty lists since you removed that check? I guess that's applicable for most changes you made on String and Collection type comparisons.

FredrikMunter: Aah, I misread `itemLines` and `serviceChargeLine`. Nevermind me.

adamvoncorswant: Check

izettle-cart/src/main/java/com/izettle/cart/CartUtils.java
```diff
@@ -407,14 +402,23 @@ static Long getServiceChargeValue(
...
             : (coalesce(serviceCharge.getAmount(), 0L) + coalesce(serviceChargeFromPercent, 0L));
     }
+
+    public static <T, S extends T> T coalesce(T subject, S fallback) {
+        return subject != null ? subject : fallback;
+    }
+
+    public static boolean empty(Collection o) {
+        return o == null || o.isEmpty();
+    }
```
softarn: This is only used in `Cart` 2 times and both those times the lists wont be null. So we can remove this method and replace the two instances with just `isEmpty`.

adamvoncorswant: You're right! BRB

adamvoncorswant:
Rebased and force pushed.

softarn:
Nice :rabbit:

adamvoncorswant:
Thanks guys!

PR #152, opened 2016-02-03
[for discussion] izettle-java: Duplicate project for compatibility

adamvoncorswant:
izettle-java-compat is intended for the clients that cannot use source/target level higher than java 1.6.

The current izettle-java will roll forward source/target to java 1.8

Note: The code is just copied directly from current izettle-java as-is. Only change is the package name (to not get class conflicts in depending repositories).

Thoughts on this, @staffanjonsson @oskarwiksten and @softarn for example? Is there a cleaner way forward?


vincenthauser:
Who will use this apart from Android then?

adamvoncorswant:
Well, no one really. But to not drag along any java8 dependencies to android, izettle-reader-library would also have to use this one as well instead of the java8 one.

adamvoncorswant:
Solved in other ways. Closing.

PR #154, opened 2016-02-11
[Do not merge] Added a mapper for DataStax's Cassandra driver that doesn't require setters

minisu:
I would like your thoughts on whether this is a worthwhile effort. This mapper builds on DataStax's mapper, but eliminates the need for mapped classes to have setters for all fields (uses a JDBI-style mapper interface instead).

Since the DataStax driver is not designed to be extended/modified, some classes has to copy a lot of code and live in the `datastax` package.

Advantages:
  - No setters required. Our objects can continue to be immutable.
  - Builds on Datastax's mapper, so it's easy to switch to that one if they improve.

Disadvantages:
  - We have to maintain this mapper.
  - The implementation is not very pretty (since Datastax's mapper is not designed to be modified).
  - It doesn't support Accessors (and possibly other things).

Ping @iZettle/partnerapps

### DIFF
--- a/datastax-cassandra-mapper/pom.xml
+++ b/datastax-cassandra-mapper/pom.xml
@@ -1,0 +1,34 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+		 xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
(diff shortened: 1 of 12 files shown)


minisu:
I'm leaning towards not using an object mapper at all. I think we can do without an object mapper and get away with the same LoC, lower complexity and fewer compromises.

PR #155, opened 2016-02-14
Izettle-java use java8

adamvoncorswant:
As we currently don't have any reasons to stay with java6 anymore.

Changed target version in pom
Switched to use diamond inference

Introduced `toZoneId` for easy access to the java8 zone object.

Also added 8 newcomers to the list of time zones: this covers all
current available zones in java8

ping @softarn

### DIFF
--- a/izettle-java/pom.xml
+++ b/izettle-java/pom.xml
@@ -15,8 +15,8 @@
             <plugin>
                 <artifactId>maven-compiler-plugin</artifactId>
                 <configuration>
-                    <source>1.6</source>
-                    <target>1.6</target>
+                    <source>1.8</source>
+                    <target>1.8</target>
                 </configuration>
             </plugin>
         </plugins>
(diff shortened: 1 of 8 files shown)


softarn:
Nice, me and georgios were thinking of the `toZoneId` thing as well. Added tests for it and changed the parsing so we first parse to `TimeZone` and then get `ZoneId` from there. There are some 3 letter abbreviations that doesn't seem to be supported for `ZoneId`. 

> For compatibility with JDK 1.1.x, some other three-letter time zone IDs (such as "PST", "CTT", "AST") are also supported. However, their use is deprecated because the same abbreviation is often used for multiple time zones (for example, "CST" could be U.S. "Central Standard Time" and "China Standard Time"), and the Java platform can then only recognize one of them.

adamvoncorswant:
@softarn : Interesting! I was actually thinking about removing the `TimeZone` compatibility completely (but realized that it might be good to do this in a second iteration). Do you think there is real value in ensuring backwards compatibility like your suggestion?

softarn:
As long as we use `TimeZone` we should have the backwards compatibility I think. But I agree, let's look into removing `TimeZone` in a second iteration. And if we remove `TimeZone` remove the three-letter time zone IDs of course.

adamvoncorswant:
Check!

PR #156, opened 2016-02-14
izettle-java instant truncator

adamvoncorswant:
`InstantTruncator`:
Unfortunately, the new java.time API doesn't allow for truncating units
larger than days, which is a show stopper for quite a few business
problems.

This utility class simply mimics the truncation behaviour of the
`CalendarTruncator` in the toolbox: just wrapping around an `Instant`
instead of a `Date`. Also, the `plus` and `minus` methods of `Instant`
also doesn't allow for units larger than a 24H, so those are added as
well for utility (so that the caller doesn't have to convert to a
`ZonedDateTime`).

ping @softarn and @aaanders : thoughts on this approach?

### DIFF
--- a/izettle-java/src/main/java/com/izettle/java/InstantTruncator.java
+++ b/izettle-java/src/main/java/com/izettle/java/InstantTruncator.java
@@ -1,0 +1,114 @@
+package com.izettle.java;
+
+import static com.izettle.java.CalendarTruncator.truncateInstant;
+
+import com.izettle.java.CalendarTruncator.CalendarField;
+import java.time.Instant;
+import java.time.ZonedDateTime;
+import java.time.temporal.ChronoUnit;
+import java.util.Date;
+
+public class InstantTruncator {
+
+    private InstantTruncator() {
+    }
+
+    /**
+     * Truncates the given instant to the nearest previous <code>ChronoUnit</code>: all lesser units are set to zero.
+     * This method acts as a complement to the standard <code>truncatedTo</code> method available on the instant. This
+     * is to be able to truncate units larger than a "day".
+     * @param instant The instant to truncate
+     * @param timeZoneId The time zone to be taken into consideration when truncating (not really needed when truncating
+     *        smaller fields, but then again: this method is not really necessary.
+     * @param chronoUnit The chrono unit to truncate to
+     * @return a newly created copy of the given instant, with the truncated value
+     */
+    public static Instant truncate(
+        final Instant instant,
+        final TimeZoneId timeZoneId,
+        final ChronoUnit chronoUnit
+    ) {
+        if (chronoUnit.getDuration().compareTo(ChronoUnit.DAYS.getDuration()) <= 0) {
+            //Up until DAYS are supported in the native functions:
+            return ZonedDateTime
+                .ofInstant(instant, timeZoneId.toZoneId())
+                .truncatedTo(chronoUnit)
+                .toInstant();
+        } else {
+            //Need to do some special handling, as these units are not supported by the native functions (but whyyy?)
+            //We loose precision below millis here, but as we're truncating we don't need them anyway
+            return truncateInstant(timeZoneId, getCalendarField(chronoUnit), Date.from(instant)).toInstant();
+        }
+    }
+
+    /**
+     * Utility method for easily adding time to an instant. This method acts as a complement to the standard
+     * <code>Instant::plus</code> method which cannot take larger chrono units than DAY.
+     * @param instant The instant to add time to
+     * @param timeZoneId The time zone to be taken into consideration when adding
+     * @param nr the number of units to add
+     * @param chronoUnit The chrono unit to add
+     * @return a newly created copy of the given instant, with the time added
+     */
+    public static Instant plus(
+        final Instant instant,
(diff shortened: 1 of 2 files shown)


softarn:
I've been reading `Instant` source code and I think I understand why it doesn't allow operations with larger units than day. Instant is only a point in time calculated by seconds from 1970-01-01T00:00:00Z . If you want to remove hours that will always be `hours * 3600`, if you want to remove days that will always be `days * 24 * 3600`. If you want to remove a month however the amount of seconds you want to remove will always vary, you can't calculate how many seconds you want to remove.

But why just not set the day in `2016-01-04T01:00:00Z` to 01? Well the Instant doesn't have any knowledge about what day it is or anything, surely it could calculate it. But that would require calendar knowledge which would basically create it to a ZoneDateTime instead. When calling `toString` on an `Instant` it will do this using `DateTimeFormatter`.

softarn:
And for the `ZonedDateTime` the truncate still doesn't take any unit larger than `Day` but there is a method called `withDayOfMonth` and `withDayOfYear`.

adamvoncorswant:
@softarn : Thanks. Yes, I totally understand the plus and minus operation on the instant object (you can only truncate on larger fields if you do have a time zone). But when it comes to `ZonedDateTime`, I don't see why they limited on DAY.

So, this utility is to be able to truncate an `Instant` directly (while of course passing in a time zone), and get an `Instant` back. In this way, the caller doesn't have to go through the hoops of converting between `LocalDateTime`/`ZonedDateTime` each time. Especially since we in almost all server-side code always handle machine time (= `Instant`).

Thoughts?

softarn:
Well `ZonedDateTime` is also storing an instant the same way Instant is and using the same calculation. Maybe for consistency or something that they decided to keep it the same way everywhere and opted for `withDayOfX` methods. Just elaborating a bit but truncating on week can perhaps be a bit weird since the week can start on different days, don't know if that depends on timezone.

However I'm not sure that we need the truncate class. The times when we need to truncate a date I think we will actually have an other date object than Instant to start with or that we want to end up with a different one. I looked through a few examples in Sun when we are using the truncate method and believe that this was true in those instances.

adamvoncorswant:
There is nothing weird in truncating on week, but of course you'll have to specify timezone (and possibly week enumeration standard as well, but that's another matter). 

So, to summarize: 
* Yes, you can probably do all things directly with the shiny java.time API, but you'll have to jump through quite a few hoops to get there (although I cannot find anything similar to truncation logic for fields greater than DAY: `withDayOfX` doesn't fit the bill. see http://www.postgresql.org/docs/9.1/static/functions-datetime.html#FUNCTIONS-DATETIME-TRUNC).
* This class mimics the old class we've been using heavily, and facilitates migration to `Instant` instead of `Date`.

My opinion is that we need this. We use `forwardInstant` and `truncateInstant` in far too many places to migrate entirely to the out-of-the-box java.time API. A middle ground, at least removing the usage of `Date`, is to enable this new truncator. In that way, we can at least switch to `Instant` without having to put at risk or alter any business logic

adamvoncorswant:
BTW: Removed the `InstantFormatter` as discussed IRL

adamvoncorswant:
@softarn : any more input on this, or are we good to go?

izettle-java/src/main/java/com/izettle/java/InstantTruncator.java
```diff
@@ -0,0 +1,114 @@
...
+     * @param instant The instant to add time to
+     * @param timeZoneId The time zone to be taken into consideration when adding
+     * @param nr the number of units to add
+     * @param chronoUnit The chrono unit to add
+     * @return a newly created copy of the given instant, with the time added
+     */
+    public static Instant plus(
+        final Instant instant,
+        final TimeZoneId timeZoneId,
+        final int nr,
```
xiaodong-izettle: Maybe use `long` here?

izettle-java/src/main/java/com/izettle/java/InstantTruncator.java
```diff
@@ -0,0 +1,114 @@
...
+     * @param instant The instant to subtract time from
+     * @param timeZoneId The time zone to be taken into consideration when subtracting
+     * @param nr the number of units to subtract
+     * @param chronoUnit The chrono unit to subtract
+     * @return a newly created copy of the given instant, with the time subtracted
+     */
+    public static Instant minus(
+        final Instant instant,
+        final TimeZoneId timeZoneId,
+        final int nr,
```
xiaodong-izettle: Maybe use `long` here also?

izettle-java/src/main/java/com/izettle/java/InstantTruncator.java
adamvoncorswant: Yeah. That's probably a good idea (as the underlying API uses that as well). BRB

adamvoncorswant:
changed parameter types to `long`. Rebased and force pushed.

xiaodong-izettle:
OK, I'll merge this one, sorry @softarn. :-)

Nice work, @adamvoncorswant. :+1: :)

adamvoncorswant:
Thanks @xiaodong-izettle :)

PR #157, opened 2016-02-14
izettle-java uuid use instant instead of date

adamvoncorswant:
Note: depends on #155 

This allows keeping the full resolution of the time information within a
UUID of type 1. Previously, limitations of the `Date` class prevented
resolution below milliseconds: now with `Instant` no time information is
lost, and a resolution down to 100ns is kept between creation and
parsing.

ping @erikryverling and @xiaodong-izettle : thoughts on this?

### DIFF
--- a/izettle-java/src/test/java/com/izettle/java/UUIDFactoryTest.java
+++ b/izettle-java/src/test/java/com/izettle/java/UUIDFactoryTest.java
@@ -2,7 +2,6 @@ package com.izettle.java;
 
 import static com.izettle.java.UUIDFactory.parse;
 import static com.izettle.java.UUIDFactory.toBase64String;
-import static java.lang.Thread.sleep;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertNotEquals;
@@ -10,7 +9,8 @@ import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertTrue;
 
 import java.nio.ByteBuffer;
-import java.util.Date;
+import java.time.Instant;
+import java.time.ZonedDateTime;
 import java.util.Random;
 import java.util.UUID;
 import org.junit.Test;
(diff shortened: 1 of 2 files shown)


izettle-java/src/test/java/com/izettle/java/UUIDFactoryTest.java
```diff
@@ -146,20 +146,21 @@ public void shouldCreateTheSameAlternativeVersionOfTheSameUUIDSuccessfully() {
...
-
-        // Assert that the UUID date is after the recent current date
-        assertTrue(uuidDate.compareTo(recentCurrentDate) > 0);
+        //Also test a fixed time:
+        final Instant instant1 = ZonedDateTime.of(1995, 3, 25, 10, 12, 15, 2345, TimeZoneId.EUROPE_STOCKHOLM.toZoneId()).toInstant();
+        final UUID uuid1 = UUIDFactory.UUID1Generator.generate(instant1);
+        final Instant uuid1Instant = UUIDFactory.getInstantFromUUID1(uuid1);
+        // Assert that the UUID instant is within 100ns of the original time
+        assertEquals(uuid1Instant.getEpochSecond(), instant1.getEpochSecond());
+        assertEquals(2300, uuid1Instant.getNano());
```
erikryverling: Shouldn't we compare `uuid1Instant` and `instant1` here as well?

erikryverling:
I think this looks great besides my small comment on the `UUIDFactoryTest`

izettle-java/src/test/java/com/izettle/java/UUIDFactoryTest.java
adamvoncorswant: Yes. I thought that it was a bit more clear if I explicitly wrote 2300 (as we're loosing 100 ns resolution), as compared to the 2345 in the original. But you're probably right, doing it more consistent with how the comparison is done on line 155 is probably better.

adamvoncorswant: Amended according to comment. Rebased and force pushed. Thanks @erikryverling !
We should still wait with merging though (PR #155 should be merged first).

adamvoncorswant:
Ok. #155 is merged. This one is good to go if you have any other objections @erikryverling 
rebased and force pushed btw, so the diff is cleaner now.

erikryverling:
erik-approves

adamvoncorswant:
Thanks @erikryverling

PR #158, opened 2016-02-14
izettle-java Interval

adamvoncorswant:
Unfortunately, the java8 time API doesn't have a proper interval class.
The only such classes in standard java is `Duration` and `Period`, but
they only hold information about the length of time: not the instant on
the timeline.
So, a proper interval class should really have been part of the standard
`java.time` API.

This class does that. It's a simple class consisting of two `Instant`s.
It's start inclusive, and end exclusive. Also, it has some utility
methods, such as `overlap`, `abuts` and `contains`.

ping @nzroller : thoughts on this? I'm thinking of quite a lot of facade methods where we're passing two `Date` objects around...

### DIFF
--- a/izettle-java/src/main/java/com/izettle/java/Interval.java
+++ b/izettle-java/src/main/java/com/izettle/java/Interval.java
@@ -1,0 +1,110 @@
+package com.izettle.java;
+
+import static java.util.Objects.requireNonNull;
+
+import java.time.Duration;
+import java.time.Instant;
+import java.util.Objects;
+import java.util.Optional;
+
+/**
+ * Represents a period on the time line between two instants. An interval will always be inclusive at the start and
+ * exclusive at the end
+ */
+public class Interval {
+
+    private final Instant start;
+    private final Instant end;
+
+    public Interval(final Instant start, final Instant end) {
+        this.start = requireNonNull(start, "Start instant cannot be null");
+        this.end = requireNonNull(end, "End instant cannot be null");
+        if (end.isBefore(start)) {
+            throw new IllegalArgumentException("An interval cannot have an end that's before it's start");
+        }
+    }
+
+    /**
+     * Specifies whether the given instant is inside this interval or not.
+     * @param instant the instant to check
+     * @return true if the instant is inside the interval, otherwise false
+     */
+    public boolean contains(final Instant instant) {
+        requireNonNull(instant, "Instant cannot be null");
+        return start.equals(instant) || (instant.isAfter(start) && instant.isBefore(end));
+    }
+
+    /**
+     * Calculates the overlap between this and the other interval, and returns the result as a new Interval
+     * @param other the other interval to check overlap against, never null
+     * @return an Optional with a present value if the two intervals overlaps, or absent otherwise. Never null.
+     */
+    public Optional<Interval> overlap(final Interval other) {
+        requireNonNull(other, "Other interval cannot be null");
+        if (disjoint(other) || abuts(other)) {
+            return Optional.empty();
+        }
+        return Optional.of(
+            new Interval(
+                start.isAfter(other.start) ? start : other.start,
+                end.isBefore(other.end) ? end : other.end
+            )
+        );
+    }
+
+    /**
+     * Figures out whether the other interval abuts this interval in either end.
+     * @param other the other interval to compare with. Cannot be null
+     * @return true if this interval ends just were the other starts, or starts where the other ends. false otherwise
+     */
+    public boolean abuts(final Interval other) {
+        requireNonNull(other, "other interval cannot be null");
+        return start.equals(other.end) || other.start.equals(end);
+    }
+
+    /**
+     * Figures out whether the other interval is completely before or completely after this interval
+     * @param other the other interval to compare with. Cannot be null
+     * @return true if the other interval is completely before or completely after this interval
+    */
+    public boolean disjoint(final Interval other) {
+        requireNonNull(other, "other interval cannot be null");
+        return start.isAfter(other.end) || other.start.isAfter(end);
+    }
+
+    public Duration getDuration() {
+        return Duration.between(start, end);
+    }
+
+    public Instant getStart() {
+        return start;
+    }
+
+    public Instant getEnd() {
+        return end;
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(start, end);
+    }
+
+    @Override
+    public boolean equals(Object obj) {
+        if (this == obj) {
+            return true;
+        }
+        if (obj == null) {
+            return false;
+        }
+        if (getClass() != obj.getClass()) {
+            return false;
+        }
+        final Interval other = (Interval) obj;
+        if (!this.start.equals(other.start)) {
+            return false;
+        }
+        return this.end.equals(other.end);
+    }
+
+}
--- a/izettle-java/src/test/java/com/izettle/java/IntervalTest.java
+++ b/izettle-java/src/test/java/com/izettle/java/IntervalTest.java
@@ -1,0 +1,91 @@
+package com.izettle.java;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+
+import java.time.Duration;
+import java.time.Instant;
+import org.junit.Test;
+
+public class IntervalTest {
+
+    @Test
+    public void itShouldFigureOutAbuts() throws Exception {
+        final Instant instant1 = Instant.parse("2007-09-01T10:15:30.00Z");
+        final Instant instant2 = Instant.parse("2013-10-02T10:15:30.00Z");
+        final Instant instant3 = Instant.parse("2015-11-03T10:15:30.00Z");
+        final Instant instant4 = Instant.parse("2017-12-03T10:15:30.00Z");
+        final Interval interval1 = new Interval(instant1, instant2);
+        final Interval interval2 = new Interval(instant2, instant3);
+        final Interval interval3 = new Interval(instant3, instant4);
+        final Interval encompassing = new Interval(instant1, instant4);
+        final Interval zeroDuration = new Interval(instant1, instant1);
+        assertTrue(interval1.abuts(interval2));
+        assertTrue(interval2.abuts(interval1));
+        assertTrue(interval2.abuts(interval3));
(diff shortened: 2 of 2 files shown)


adamvoncorswant:
Also pinging @ssprang for your thoughs :)

nzroller:
Looks good @adamvoncorswant!

ssprang:
Yes, looks good to me too! Nice one, is there a pr coming up that is using this?
Before we merge perhaps @chas678 could have a look too?

izettle-java/src/main/java/com/izettle/java/Interval.java
```diff
@@ -0,0 +1,112 @@
...
+    public Instant getStart() {
+        return start;
+    }
+
+    public Instant getEnd() {
+        return end;
+    }
+
+    @Override
+    public int hashCode() {
```
chas678: Objects.hash is simpler here: http://docs.oracle.com/javase/8/docs/api/java/util/Objects.html#hash-java.lang.Object...-

izettle-java/src/main/java/com/izettle/java/Interval.java
```diff
@@ -0,0 +1,112 @@
...
+        }
+    }
+
+    /**
+     * Specifies whether the given instant is inside this interval or not.
+     * @param instant the instant to check
+     * @return true if the instant is inside the interval, otherwise false
+     */
+    public boolean contains(final Instant instant) {
+        assertNotNull(instant, "Instant cannot be null");
```
chas678: Bespoke method instead of Objects.requireNonNull: http://docs.oracle.com/javase/8/docs/api/java/util/Objects.html#requireNonNull-T-java.lang.String-

izettle-java/src/main/java/com/izettle/java/Interval.java
```diff
@@ -0,0 +1,112 @@
...
+            return true;
+        }
+        if (obj == null) {
+            return false;
+        }
+        if (getClass() != obj.getClass()) {
+            return false;
+        }
+        final Interval other = (Interval) obj;
+        if (!this.start.equals(other.start)) {
```
chas678: Objects.equals(a,b) ? - http://docs.oracle.com/javase/8/docs/api/java/util/Objects.html#equals-java.lang.Object-java.lang.Object-

chas678:
I don’t really understand the use-case/premise of this class TBH -  I’m guessing there is some reason not to just use something like Guava Ranges, JodaTime or threeten-extra extensions to java.time?  

* https://github.com/ThreeTen/threeten-extra
* http://joda-time.sourceforge.net/apidocs/org/joda/time/Interval.html
* http://joda-time.sourceforge.net/key_interval.html
* https://github.com/google/guava/wiki/RangesExplained

izettle-java/src/main/java/com/izettle/java/CollectionUtils.java
```diff
@@ -36,7 +36,7 @@ private CollectionUtils() {
         if (largeCollection == null) {
             return null;
         }
-        List<Collection<T>> retList = new LinkedList<Collection<T>>();
+        List<Collection<T>> retList = new LinkedList<>();
```
chas678: Is this method necessary now?  Can we just compose a call to  Guava (it's on the dependency list for toolbox) - and it seems it can already flexibly partition collections using `Multimaps index` :  http://docs.guava-libraries.googlecode.com/git/javadoc/com/google/common/collect/Multimaps.html#index%28java.lang.Iterable,%20com.google.common.base.Function%29
Use Guava's Multimaps.index.

Here is an example, which partitions a list of words into two parts: those which have length > 3 and those that don't.
```java
List<String> words = Arrays.asList("foo", "bar", "hello", "world");

ImmutableListMultimap<Boolean, String> partitionedMap = Multimaps.index(words, new Function<String, Boolean>(){
    @Override
    public Boolean apply(String input) {
        return input.length() > 3;
    }
});
System.out.println(partitionedMap);
prints:

false=[foo, bar], true=[hello, world]
```

And if you already have a pre=dicate for the split you could turn it into a Function with `Functions.forPredicate`

izettle-java/src/main/java/com/izettle/java/CollectionUtils.java
```diff
@@ -46,11 +46,11 @@ private CollectionUtils() {
...
-                    part = new TreeSet<T>();
+                    part = new TreeSet<>();
                 } else if (largeCollection instanceof Set) {
-                    part = new HashSet<T>();
+                    part = new HashSet<>();
                 } else if (largeCollection instanceof List) {
-                    part = new LinkedList<T>();
+                    part = new LinkedList<>();
                 } else {
                     throw new UnsupportedOperationException("Not prepared for this type of collection: " + largeCollection.getClass().toString());
```
chas678: Unnecessary 'toString()' call at line 55

izettle-java/src/main/java/com/izettle/java/TimeZoneId.java
```diff
@@ -623,6 +624,14 @@
     PACIFIC_ENDERBURY("Pacific/Enderbury"),
```
chas678: private modifier is redundant for Enum constructors

izettle-java/src/main/java/com/izettle/java/Interval.java
adamvoncorswant: Nice. Thanks

izettle-java/src/main/java/com/izettle/java/Interval.java
adamvoncorswant: I reasoned that as none of the members can be null, so the benefit of the null safe `Objects.equals` is not needed here, and that this is clearer. Don't you agree?

izettle-java/src/main/java/com/izettle/java/Interval.java
adamvoncorswant: Well, we *do* have the `assertNotNull` method in the `ValueChecks` class. I'm up for using either, but I thought that I should use the new additions by @pgisslen instead of the ones in `Objects`. Mostly because I like the `IllegalArgumentException` better in this context than `NullPointerException`. I'm not religious over this though, so just holler if you want me to change.

izettle-java/src/main/java/com/izettle/java/CollectionUtils.java
adamvoncorswant: 1. This PR is actually only about the added `Interval` class. The above is just an accidental diff until PR #155 is merged
2. I would argue that the question whether we should have the method or not is outside even of PR #155, as the code is just prettified with diamond operator since we now switched to java8
3. We don't have any dependencies at all here, so guava is not an option.

izettle-java/src/main/java/com/izettle/java/CollectionUtils.java
adamvoncorswant: Same here. Out of scope for this PR?

izettle-java/src/main/java/com/izettle/java/TimeZoneId.java
adamvoncorswant: I don't understand what you mean. Care to clarify?

adamvoncorswant: Duh. Never mind. I'll fix this in PR #155

izettle-java/src/main/java/com/izettle/java/Interval.java
chas678: OK.  maybe later @pgisslen can talk about why this was re-implemented not reused.
This functionality is in both Guava Preconditions and java.util.Objects both of which are on the classpath for toolbox.

A better API might have been to copy the exceptions thrown by Guava Preconditions (NPE for Null and IAE for Arguments) .. but then we could just use Guava: https://github.com/google/guava/wiki/PreconditionsExplained
ValueChecks.assertNotNull > IllegalArgumentException is a bit strange - as it assumes it's always an argument, but, as you say.. this one actually *is*.  
And I suppose, philosophically, it is *always* an argument to `assertNotNull` :-)

adamvoncorswant:
@chas678 : Regarding your PR-wide questions:
To me this is the class that we should have used in basically all method signatures where we currently have a `Date from, Date to` (or `Interval from, Interval to` of course). Unfortunately, we haven't done that as of yet, but I want to be able to change the existing signatures.
The reason to why we don't use joda-time, is because we don't use joda-time in favour of java8 time.  This class is introduced in the toolbox, because we use the toolbox everywhere, and it doesn't introduce any additional dependencies (which is also why we don't want to use guava or three-ten-extra either).

Also, I believe that `Interval` will be part of java9, so this is also to prepare for an easy switch when that time comes.

izettle-java/src/main/java/com/izettle/java/Interval.java
softarn: Yea we need to decide how we check for null. I find going against, what I believe is, the Java standard of checking for null with `Objects.requireNotNull` and not throwing `NullPointerException` strange. But I suppose that's for an other discussion. Feel free to include me if you need input to one.

adamvoncorswant: Aight. I'm switching back to `Objects.requireNonNull` then :)

izettle-java/src/main/java/com/izettle/java/Interval.java
chas678: sure, sounds good.  BTW - there is a nice plugin for intellij that generates equals/hashcode using choice coders choice of Guava or java.util.Objects that I think is useful - more modern code generation than the intellij built in IDEA generator - it's called EqualsHashCodeDeluxeGenerator : https://github.com/mjedynak/EqualsHashCodeDeluxeGenerator

adamvoncorswant: Nice stuff @chas678 !

Side note: Netbeans (which I believe I'm the only user of) actually does use `Objects` by default. In this specific case I changed it back to "old style" versions because of the above reasoning. Perhaps it's better to always use `Objects`. What do I know :)

adamvoncorswant:
Anyone of you guys have any more thoughts on this, or are we good to go?

xiaodong-izettle:
I think this is good to go. Nice work, @adamvoncorswant. :+1: :)

adamvoncorswant:
Thanks @xiaodong-izettle :)

PR #159, opened 2016-02-17
Use try-with-resources idiom on autocloseables

chas678:
Findbugs reports a number of resourceleaks where resources are not being closed on every execution path.
Adding try-with-resources to those it reports to stop resource leaks.

/cc @mxns + @jakanjakan

### DIFF
--- a/izettle-java/src/main/java/com/izettle/java/ResourceUtils.java
+++ b/izettle-java/src/main/java/com/izettle/java/ResourceUtils.java
@@ -30,7 +30,7 @@ public abstract class ResourceUtils {
      * @throws IOException On resource not found
      */
     public static byte[] getResourceAsBytes(String resourceName)
-            throws IOException {
+        throws IOException {
         ResourceLoader resourceLoader = new ResourceLoader();
         return resourceLoader.getBytesFromResource(resourceName);
     }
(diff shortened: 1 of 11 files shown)


izettle-messaging/src/main/java/com/izettle/messaging/handler/MeteredMessageHandler.java
```diff
@@ -27,9 +27,7 @@ public MeteredMessageHandler(MessageHandler<M> actualHandler, MetricRegistry met
 
     @Override
     public void handle(M message) throws Exception {
-
-        Timer.Context timerContext = timer.time();
-        try {
+        try (Timer.Context timerContext = timer.time()) {
```
chas678: works as close, (via autocloseable), is equivalent to stop - https://dropwizard.github.io/metrics/3.1.0/apidocs/com/codahale/metrics/Timer.Context.html

izettle-java/src/main/java/com/izettle/java/ResourceUtils.java
```diff
@@ -115,21 +116,18 @@ private static String removePreSlash(String resourceName) {
...
+        try (ByteArrayOutputStream outputStream = new ByteArrayOutputStream()) {
             byte[] buffer = new byte[1024];
             int read;
             while ((read = inputStream.read(buffer)) != -1) {
                 outputStream.write(buffer, 0, read);
             }
-            outputBytes = outputStream.toByteArray();
+            return outputStream.toByteArray();
         } finally {
             if (inputStream != null) {
```
ssprang: Hmm.. Weird to take the inputstream as a parameter here and then close the stream..

PR #160, opened 2016-02-18
Add unit test for SqlArrayFactory

chas678:
Unit test was missing for the build method.

/cc @mxns

### DIFF
--- a/izettle-jdbi/src/main/java/com/izettle/jdbi/PostgresArrayArgumentFactory.java
+++ b/izettle-jdbi/src/main/java/com/izettle/jdbi/PostgresArrayArgumentFactory.java
@@ -1,7 +1,6 @@
 package com.izettle.jdbi;
 
 import java.sql.Array;
-import java.sql.Connection;
 import org.skife.jdbi.v2.StatementContext;
 import org.skife.jdbi.v2.tweak.Argument;
 import org.skife.jdbi.v2.tweak.ArgumentFactory;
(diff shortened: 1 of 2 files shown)


ssprang:
Nice with a test for this :)
I like tests to include extremes though.. What happens (what do we expect to happen) when 1 vs 0 arguments are present in the array?

izettle-jdbi/src/main/java/com/izettle/jdbi/PostgresArrayArgumentFactory.java
```diff
@@ -1,7 +1,6 @@
 package com.izettle.jdbi;
 
 import java.sql.Array;
-import java.sql.Connection;
```
ssprang: Hehe, just noticed that this fails checkstyle in TC master now, weird that it was green on the PR though...

chas678:
@ssprang - ref test edge cases - Have added more unit test to the build method for each of the supported types.
Edge cases of delinquent inputs (should be) handled by the PreparedStatement class implementation in the JDBC driver - params count from 1 (not zero) and the JDBC spec says it should throw if no param exists at the offset referenced, but this is driver dependent and not tested.  These are the exception a driver is supposed to throw on `setArray` http://docs.oracle.com/javase/8/docs/api/java/sql/PreparedStatement.html#setArray-int-java.sql.Array-

izettle-jdbi/src/main/java/com/izettle/jdbi/PostgresArrayArgumentFactory.java
chas678: </hangs head in shame> I know why but am not telling....

PR #162, opened 2016-02-24
Removed truncate method from InstantTruncator and created InstantTrun…

softarn:
…catoeTemporalAdjuster

Ping @adamvoncorswant and @xiaodong-izettle 

I missed the last PR but I've read up a bit more and think that TemporalAdjusters is the way to go if we want to truncate instants. (Even though I'm not sure that we will need to truncate instances in this way but let's see).

I skipped supplying the zoneId and using UTC since Instant is in UTC and setting that weeks start on Monday as in CalendarTruncator.


softarn:
https://docs.oracle.com/javase/tutorial/datetime/iso/adjusters.html

xiaodong-izettle:
:+1:

adamvoncorswant:
I'm sorry @softarn , but I don't think this makes any sense. Why would you want to truncate on UTC? No one lives in UTC: it's a virtual time zon (I'd actually go so far to say that it's wrong by the API's to allow you to talk about calendar fields together with UTC, but that's another issue :). The *entire purpose* of the existing `InstantTruncator` implementation is to allow truncation of an `Instant` in any given time zone. 

It's fully possible that the TemporalAdjuster is a really good class, but as far as I see you cannot provide time zone in the call, which would assume that the tz information is included in the `temporal` (for example a ZonedDateTime) but this defeats the purpose of the current implementation.

xiaodong-izettle:
IMHO, I think this interface `TemporalAdjuster` is good stuff. But as @adamvoncorswant said, we should not make assumption on `UTC`. By the way, @adamvoncorswant, do the returning of `TemporalAdjusters::firstDayOfMonth` and `TemporalAdjusters::firstDayOfNextMonth` suit our needs?

adamvoncorswant:
@xiaodong-izettle : quite possible. If so, we should in my mind use those to remove the dependency on `CalendarTruncator`, while still keeping the public methods in `InstantTruncator` unaffected.

softarn:
I could add zoneId to the constructor and work with that as well.

softarn:
I've added zoneId to constructor.

softarn:
If you have a timezone and an instant you should be using the ZonedDateTime object instead of having an instant and the timezone as two separate variables and jumping between ZonedDateTime and Instant everytime you want to do some work on it.

adamvoncorswant:
Well, that depends on how you currently use the api, and how you store your timestamps in the persistence layer, right? This is probably all good @softarn , but the current ambition for me is not to rewrite all current logic in other systems, but rather to have a viable drop in replacement from `Date` to `Instant`.

If you still like to have this incorporated, I'm all go.

softarn:
My concern is just code quality and using the apis in a correct way. Since Java 8 have set this way of doing adjustments I think it is a good practise to do so.

If you only want a class for doing the migration from `Date` to `Instant`, even though it in many cases would be better to use `ZonedDateTime` or an other class, then I suggest that we put a `@Deprecated` annotation on it directly. Because I don't believe that truncating instances on units larger than day and in a certain timezone will be a common pattern. It is possible that we find out that it is needed but let's add it when we do know that it is the case.

Since this migration only applies to some repositories I don't think this truncating class should be in this repository but rather be put in those repositories that need them.

But if I choose between having this class and the `InstantTruncator` I choose this.

softarn:
@adamvoncorswant updated with better readability

```
instant.with(truncationBy(ChronoUnit.MINUTES, zoneId));
```

izettle-java/src/main/java/com/izettle/java/InstantTruncateAdjusters.java
```diff
@@ -0,0 +1,47 @@
...
+ * Truncates an instant to the supplied ChronoUnit.<br>
+ * <br>
+ * For example creating an instance with {@link java.time.temporal.ChronoUnit#HOURS} and using it to adjust the
+ * instant 2016-02-24T12:12:55.854Z will adjust the instant into 2016-02-24T12:00:00.000Z <br>
+ * <br>
+ * Adjusting an instant weeks will use the ISO-8601 standard where weeks start on Mondays
+ */
+public class InstantTruncateAdjusters {
+
+    public static TemporalAdjuster truncationBy(ChronoUnit chronoUnit, ZoneId zoneId) {
```
adamvoncorswant: Nice! 

When we now have a method that called `truncationBy`: shouldn't we remove that part of the class name? I mean, we might possibly want to add more adjusters to this class later on, that has nothing to do with truncations. So, I'd suggest just `InstantAdjusters` as a class name (much in the same way as java.utils.stream.Collectors has multiple methods with entirely different purpose).

izettle-java/src/main/java/com/izettle/java/InstantTruncateAdjusters.java
```diff
@@ -0,0 +1,47 @@
...
+        return (temporal) -> {
+            ZonedDateTime zonedDateTime = ZonedDateTime.ofInstant(Instant.from(temporal), zoneId);
+
+            if (chronoUnit.compareTo(ChronoUnit.DAYS) <= 0) {
+                zonedDateTime = zonedDateTime.truncatedTo(chronoUnit);
+            } else {
+                zonedDateTime = zonedDateTime.truncatedTo(ChronoUnit.DAYS);
```
adamvoncorswant: What do you think about being more explicit here to actually support all units? I'm thinking either switching over all possible values (and return directly from each case, or at least throw exception when the chronoUnit is not supported. Such as MILLENNIA and FOREVER for example.

softarn: Ah yea lost the exception when refactoring. Added back with test.

adamvoncorswant:
Great @softarn : you're feel ready with this? Want me to merge?

softarn:
go!

adamvoncorswant:
:+1:

PR #164, opened 2016-03-02
Add AwsSecurityGroupHostSupplier support for Datastax driver

karlgrund:
Added support for Datastax driver for AwsSecurityGroupHostSupplier

* Added interface to return ConnectionPoint which then is converted to suit either Astyanax or Datastax driver.

**No tests written**

@iZettle/partnerapps ping!


karlgrund:
Mikael, could you take a look at this? Basically a refactor to work on Datastax as well.

izettle-cassandra/src/main/java/com/izettle/datastax/impl/AwsSecurityGroupHostSupplier.java
```diff
@@ -0,0 +1,53 @@
...
+import com.amazonaws.services.ec2.AmazonEC2Client;
+import com.izettle.cassandra.AwsSecurityGroupHostSupplierAbstract;
+import com.izettle.cassandra.ConnectionPoint;
+import java.net.InetSocketAddress;
+import java.util.Collection;
+import java.util.List;
+import java.util.stream.Collectors;
+
+/**
+ * Cassandra Astyanax Host Supplier for Multiregion AWS implementations,
```
mikaellothman: You could probably be a bit more clear on which implementation that should be used for which driver (Astyanax or Datastax)?

karlgrund: I was thinking about renaming the methods, but wanted to prevent those using Astyanax from having to alter their current implementation.

mikaellothman: I understand, but shouldn't at least the javadoc reflect that..

minisu:
Looks great!

A few things though:

Split into two modules to avoid dragging in Astyanax/DataStax when not needed?

Not naming the two impl classes the same name?

"Abstract" being first in the class name instead of last?

karlgrund:
ping @minisu @mikaellothman

mikaellothman:
Looks good to me @karlgrund :+1:  Has this been tested in an environment using security group supplier?

PR #165, opened 2016-03-03
Consider `Optional.empty()` to be `ValueChecks.empty()`

oskarwiksten:
* Makes `ValueChecks.empty` return `true` when a value of `Optional.empty()` is passed as argument.


oskarwiksten:
Not really sure whom to assign this to. Anyone at @izettle-tech ?
Could this change lead to unexpected situations in existing code?

adamvoncorswant:
Seems like the only logical choice here. Nice!

adamvoncorswant:
:+1:

PR #166, opened 2016-03-27
InstantAdjusters: Moves `plus` and `minus` to adjuster class

adamvoncorswant:
And removed `InstantTruncator` completely

Naming of these methods is somewhat tricky: I choose to go with `additionBy` and `subtractionBy` (instead of just plus and minus) out of two reasons:
* More alignment with the already existing `truncationBy`
* More unique name, which makes static imports elsewhere of these methods more easy to read.

Thoughts @softarn ?


softarn:
Like the naming, looks good! :accept:

adamvoncorswant:
Thanks @softarn

PR #171, opened 2016-05-17
Increase AWS client retry times from 3 (default) to 6.

xiaodong-izettle:
For discussion: let's increase the retry time from 3 to 6, see if it helps? 

I searched for `MessagingManager NOT INFO` in Splunk for last 30 days, no results...

Ping @nzroller and @fredrikbackstrom, thanks...

### DIFF
--- a/izettle-messaging/src/main/java/com/izettle/messaging/AmazonSNSClientFactory.java
+++ b/izettle-messaging/src/main/java/com/izettle/messaging/AmazonSNSClientFactory.java
@@ -1,13 +1,21 @@
 package com.izettle.messaging;
 
+import com.amazonaws.ClientConfiguration;
+import com.amazonaws.ClientConfigurationFactory;
 import com.amazonaws.auth.AWSCredentials;
 import com.amazonaws.services.sns.AmazonSNSAsync;
 import com.amazonaws.services.sns.AmazonSNSAsyncClient;
+import java.util.concurrent.Executors;
 
 /**
  * Factory that creates Amazon SNS clients for specific endpoints.
  */
 public class AmazonSNSClientFactory {
+
+    /**
+     * Client configuration factory providing ClientConfigurations tailored to this client
+     */
+    private static final ClientConfigurationFactory CONFIG_FACTORY = new ClientConfigurationFactory();
 
     /**
      * Creates Amazon SNS client for given endpoint using the provided credentials.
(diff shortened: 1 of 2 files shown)


fredrikbackstrom:
ping @AndreasMeisingseth

fredrikbackstrom:
@xiaodong-izettle then the failure is not logged since we are not handling the Future.error in Sun. 

I think it's worth doing this as it would at least limit the effect of network problems. It's a aid and let's find more solutions further on. 

I wonder though how this can be the root cause of the missing events problems as the default 3 would catch most.

xiaodong-izettle:
@fredrikbackstrom, thanks for the comment. I assume the container would log for `EJBException` anyway... 

But just in case, I also check this in Splunk `AmazonHttpClient NOT INFO NOT "Unable to parse SQS queue URL" NOT host="batch-processing2" NOT host="riskapp1-ie-a" NOT "Will leave it on queue"` for last 30 days. 15 results, 8 from `app2-sp-b`, which should be fine, as Planet has a build-in solution, 7 from `izlvmdapp001`, which kinds of understandable, as that server is not ec2 instance...

AndreasMeisingseth:
Nice!

xiaodong-izettle:
And I agree with you, @fredrikbackstrom, that retrying 3 times should catch most.... 

The mismatch due to SQS/SNS should be very very few....

xiaodong-izettle:
So @AndreasMeisingseth, it would be very helpful that, you could provide us some examples of the issue you mentioned, so that we could dig deep, to find the root cause... :)

AndreasMeisingseth:
@xiaodong-izettle I'm preparing that for you, will send it over shortly (like tomorrow or so)

xiaodong-izettle:
Thanks a lot in advance, @AndreasMeisingseth. :)

xiaodong-izettle:
@AndreasMeisingseth, any updates?

AndreasMeisingseth:
The updates I have is that it looks better than before (I've only looked at data from 2016-02-01 going forward) when we haven't experienced any SNS/SQS outages. We are only having single events missing, which is within the tolerance of what we can tolerate.

AndreasMeisingseth:
With that said, I still think that this is the way forward, and in the event of time add a mechanism that makes us resilient to AWS SNS/SQS going down for longer periods (+1 minute). We have encountered that before, and will most likely do that again.

xiaodong-izettle:
OK, thanks for the update, @AndreasMeisingseth. :)

fredrikbackstrom:
So the move of SUN to AWS makes network communication to SNS/SQS more reliable. 

Then shall we still try this out or should we skip this and focus on solving the issue when the downtime is > 1 min.

xiaodong-izettle:
IMHO, we just keep what we have for now... :)

I'll close this one, but feel free to open, if anyone thinks this is still what we need...

PR #173, opened 2016-06-21
Java8ify interval

softarn:
Since this is closely related to the java time package I think we should follow their convention with `of` methods for instantiation.

### DIFF
--- a/izettle-java/src/main/java/com/izettle/java/Interval.java
+++ b/izettle-java/src/main/java/com/izettle/java/Interval.java
@@ -16,7 +16,11 @@ public class Interval {
     private final Instant start;
     private final Instant end;
 
-    public Interval(final Instant start, final Instant end) {
+    public static Interval of(final Instant start, final Instant end) {
+        return new Interval(start, end);
+    }
+
+    private Interval(final Instant start, final Instant end) {
         this.start = requireNonNull(start, "Start instant cannot be null");
         this.end = requireNonNull(end, "End instant cannot be null");
         if (end.isBefore(start)) {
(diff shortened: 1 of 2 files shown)


adamvoncorswant:
Any reason for now allowing the public constructor in parallell?

softarn:
There is no public constructor with this PR

adamvoncorswant:
That's exactly my question: why don't we allow that as well?

softarn:
Ah I see, not sure why they went for this convention. Let's wait for my question a bit: http://programmers.stackexchange.com/questions/322936/why-do-java-time-have-methods-for-creating-objects-instead-of-just-constructors

adamvoncorswant:
Cool. Let's wait then :)

softarn:
Answers are in. So they basically went for them to be able to be in control of which instance to return. For example the `Month` object which returns the enum from the private `ENUMS` cache

```
    public static Month of(int month) {
        if (month < 1 || month > 12) {
            throw new DateTimeException("Invalid value for MonthOfYear: " + month);
        }
        return ENUMS[month - 1];
    }
```

And descriptive power of calling methods i.e. `ofInstant`.

So I don't see that these will apply to `Interval`. I do think we should have the constructor private to only have one way of creating the object and that we should go for `of` because it's the standard of the `java.time` package which this is building on.

adamvoncorswant:
Aight. So, do you think we should merge this as is then?

softarn:
Yes, thanks! I'll make sure to update in the sun.

adamvoncorswant:
Thanks @softarn

PR #175, opened 2016-07-05
Also remove SNS envelope from msgs produced by `wrapInSNSMessage`

oskarwiksten:
* Makes the method `MessageDeserializer::removeSnsEnvelope` also remove SNS envelopes from messages that are produced by `QueueServiceSender::wrapInSNSMessage`.
* With this change, messages that are produced by either "regular" SNS publishing, directly to SQS by using `MessagePublisher`, and raw messages sent by `MessageQueueProducer`, will now all be correctly deserialized when being processed by a `MessageHandler`. Previously, messages sent directly to SQS by using `MessagePublisher` would not have their SNS envelopes removed when being handled by a `QueueServicePoller`.

### DIFF
--- a/izettle-messaging/src/main/java/com/izettle/messaging/serialization/MessageDeserializer.java
+++ b/izettle-messaging/src/main/java/com/izettle/messaging/serialization/MessageDeserializer.java
@@ -46,7 +46,7 @@ public class MessageDeserializer<M> {
     public static String removeSnsEnvelope(String message) throws IOException {
         if (!empty(message) && message.startsWith("{")) {
             JsonNode root = JSON_MAPPER.readTree(message);
-            if (root.has("TopicArn") && root.has("Message")) {
+            if (root.has("Subject") && root.has("Message")) {
                 return root.get("Message").asText();
             }
         }
--- a/izettle-messaging/src/test/java/com/izettle/messaging/TestMessage.java
+++ b/izettle-messaging/src/test/java/com/izettle/messaging/TestMessage.java
@@ -1,5 +1,8 @@
 package com.izettle.messaging;
 
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+
+@JsonIgnoreProperties(ignoreUnknown = true)
 public class TestMessage {
     private String message;
 
--- a/izettle-messaging/src/test/java/com/izettle/messaging/serialization/MessageDeserializerTest.java
+++ b/izettle-messaging/src/test/java/com/izettle/messaging/serialization/MessageDeserializerTest.java
@@ -1,9 +1,9 @@
 package com.izettle.messaging.serialization;
 
+import static com.izettle.java.ResourceUtils.getResourceAsBytes;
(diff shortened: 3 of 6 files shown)


oskarwiksten:
Ping @xiaodong-izettle or even @mikaellothman since this is a public repo.

xiaodong-izettle:
😀 👍

izettle-messaging/src/main/java/com/izettle/messaging/serialization/MessageDeserializer.java
```diff
@@ -46,7 +46,7 @@ public M deserialize(String message) throws IOException {
     public static String removeSnsEnvelope(String message) throws IOException {
```
xiaodong-izettle: @oskarwiksten, should we rename this method? 

If we are afraid of break backward compatibility, we could create a new method, use the new method, and deprecate this one...

oskarwiksten: I think the name correctly states the intention of what this method does (it strips the json properties that SNS adds, and returns the contents of the actual message that was originally sent to SNS). Do you think it can be misinterpreted for something else?

I do not think it breaks backwards compatibility, at least not for iZettle's code. We never call this method directly in any of our services, and the messages that we serialize never use PascalCase.

oskarwiksten: If we do not want to change this deserializer, I guess another change that would accomplish the same goal would be to add code in `QueueServiceSender::wrapInSNSMessage` that adds a json property named `TopicArn`. Do you think that would be safer? (I still think this change is safe)

xiaodong-izettle: Sorry for the misunderstanding... :-)

We have `example-message-sent-with-messagepublisher-to-sqs.json`, and we call `removeSnsEnvelope` method, yes?

xiaodong-izettle: Regarding the backward compatibility, I definitely agree with you, this won't break anything for iZettle's code... Maybe that's good enough... :-)

oskarwiksten: Correct, `example-message-sent-with-messagepublisher-to-sqs.json` was produced by an instance of a `MessagePublisher`, and its message is somewhat similar to a message produces by regular SNS (since it was passed through `QueueServiceSender::wrapInSNSMessage`). This message has the "envelope" of a SNS message, in that it has a `Subject` stating the type of message, and the actual message json is contained in the `Message` property, just like SNS messages have.

xiaodong-izettle: OK, thanks for the explanation... :-)

xiaodong-izettle:
Nice one, @oskarwiksten. 👍

PR #176, opened 2016-07-06
Add factory method for creating encrypted MessagePublisher

oskarwiksten:
* Adds factory method in `QueueServiceSender` to create an encrypted version of a `MessagePublisher`.
* Reason for adding is that all of the support is already in place for encrypting messages when sending through a `MessagePublisher`, similar to what is available when using a `MessageQueueProducer<T>`.
* Long-term plan with this addition is to deprecate `MessageQueueProducer<T>` and instead only use `MessagePublisher`, since they are so similar and share almost the same responsibility. It is not clear for all callers what the difference between these two classes are, so having just one would simplify for callers.


oskarwiksten:
@xiaodong-izettle

xiaodong-izettle:
Yes, totally agree... :-)

Nice one, @oskarwiksten. 👍

PR #177, opened 2016-07-09
Fix how line item discounts are calculated

adamvoncorswant:
Line item discounts are now calculated on lines rounded gross value
instead of it's non-rounded value. This is to avoid situations where the
discount value is calculated on an exact value, and then rounded in a
direction that differs from how the final line items value are rounded:
effectively disproportionately changing the percentage of the discount
in relation to the gross amount.

Added tests that fail without the actual logical change.

ping @staffanjonsson : thoughts?


ssprang:
Nice! Thank you! 👍

adamvoncorswant:
Involved @ssprang as she discovered the bug and is closely monitoring these problems in prod

adamvoncorswant:
Thanks @ssprang !

@staffanjonsson : still interested in you thoughts around this :)

PR #178, opened 2016-07-14
Check for handlers matching Type if no Subject set

powion:
MessageDispatcher will now check for handlers matching the Type value if no Subject is set on a
received message. This is useful for Amazon notifications as they lack Subject.


freddd:
👍

PR #179, opened 2016-07-19
Add AsyncMessageDispatcher for asynchronous interaction with SQS

powion:
This dispatcher passes along the receipt handle from SQS, allowing the receiving handler to delete the message at a later point.

### DIFF
--- a/izettle-messaging/src/main/java/com/izettle/messaging/handler/AsyncMessageDispatcher.java
+++ b/izettle-messaging/src/main/java/com/izettle/messaging/handler/AsyncMessageDispatcher.java
@@ -1,0 +1,132 @@
+package com.izettle.messaging.handler;
+
+import static com.izettle.java.ValueChecks.empty;
+
+import com.amazonaws.services.sqs.model.Message;
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+import com.izettle.messaging.MessagingException;
+import com.izettle.messaging.serialization.AmazonSNSMessage;
+import com.izettle.messaging.serialization.AmazonSQSMessage;
+import com.izettle.messaging.serialization.JsonSerializer;
+import com.izettle.messaging.serialization.MessageDeserializer;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+
+/**
+ * Routes messages to other MessageHandlers based on message type. All handlers added with the
+ * addHandler() method will be called whenever a message of the supplied type is received by this class.
+ * The handlers need to delete the messages from the queue after handling them, using the receipt handle in
+ * the AmazonSQSMessage.
+ *
+ * The received message will need either a Subject or a Type field in the body, as the value of one of these fields
+ * is used to map to a handler.
+ */
+public class AsyncMessageDispatcher implements MessageHandler<Message> {
+    private final MessageDeserializer<String> messageDeserializer;
+    private final Map<String, ListOfMessageHandlersForType> messageHandlersPerEventName = new ConcurrentHashMap<>();
+    private final List<MessageHandler<AmazonSQSMessage>> defaultMessageHandlers = new ArrayList<>();
+    private static final ObjectMapper JSON_MAPPER = JsonSerializer.getInstance();
+
+    public static AsyncMessageDispatcher nonEncryptedMessageDispatcher() {
+        return new AsyncMessageDispatcher();
+    }
+
+    public static AsyncMessageDispatcher encryptedMessageDispatcher(
+        byte[] privatePgpKey,
+        final String privatePgpKeyPassphrase
+    ) throws MessagingException {
+        if (empty(privatePgpKey) || empty(privatePgpKeyPassphrase)) {
+            throw new MessagingException(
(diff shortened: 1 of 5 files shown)


freddd:
Would be good with some additional comments as well, ping @adamvoncorswant @xiaodong-izettle, looks good?

xiaodong-izettle:
Any scenario we need this?

izettle-messaging/src/main/java/com/izettle/messaging/handler/AsyncMessageDispatcher.java
```diff
@@ -0,0 +1,132 @@
...
+/**
+ * Routes messages to other MessageHandlers based on message type. All handlers added with the
+ * addHandler() method will be called whenever a message of the supplied type is received by this class.
+ * The handlers need to delete the messages from the queue after handling them, using the receipt handle in
+ * the AmazonSQSMessage.
+ *
+ * The received message will need either a Subject or a Type field in the body, as the value of one of these fields
+ * is used to map to a handler.
+ */
+public class AsyncMessageDispatcher implements MessageHandler<Message> {
```
xiaodong-izettle: This is not about `Async`, yes? It's about passing along the `receiptHandle`, so that we could delete the message later, yes?

freddd:
Yes, in the new email service we pick an item from the queue and handle it without persisting it. We don't want to remove it from the queue (only reserve it) until after we have gotten a confirmation from the SES that the email was sent (we use the async client) because we don't want to end up in a situation where we picked an item from the queue (and removed it like we usually do), the app crashed and we are not able to recover and send it again. The idea is that if we were to add two more instances of this service, both of them could start reading from the queue directly and independently of each other and start sending emails.

We previously solved this by basically storing all of the events in a db table and having another process picking them up and sending them, however this does not scale as well and could be a subject of race conditions or locking of db tables if you add more processes doing this.

I hope that explains it somehow at least :)

xiaodong-izettle:
OK, how would you solve the event duplication issue? Will multiple emails be sent out?

freddd:
As I understand it, if you reserve a message another process cannot get it until the reservation has expired which is after 30 seconds using default settings. If we are not able to send an email and get a response within 30 seconds after picking it off the queue we should probably either increase the reserve period or scale up by adding more instances.

xiaodong-izettle:
Things like this does happen in prod, you have to handle that... That's why we always use some kinds of UUID to remove the event duplication...

freddd:
Ah, I may have misunderstood you. If there are two equal events on the queue, the email service would check against those already sent to ensure that there is no such uuid that has already been delivered.

izettle-messaging/src/main/java/com/izettle/messaging/serialization/AmazonSQSMessage.java
```diff
@@ -0,0 +1,47 @@
...
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import java.util.UUID;
+
+@JsonIgnoreProperties(ignoreUnknown = true)
+@JsonInclude(JsonInclude.Include.NON_NULL)
+public class AmazonSQSMessage {
+
+    @JsonProperty("Body")
+    private AmazonSNSMessage body;
```
adamvoncorswant: I'm not all too familiar with this area, but to me it looks strange that there is an `AmazonSNSMessage` inside an `AmazonSQSMessage`. Is that really something that's a general truth (remember that this repo is open source and should be completely void of all domain specific assumptions)?

izettle-messaging/src/main/java/com/izettle/messaging/handler/AsyncMessageDispatcher.java
```diff
@@ -0,0 +1,132 @@
...
+
+    public static AsyncMessageDispatcher nonEncryptedMessageDispatcher() {
+        return new AsyncMessageDispatcher();
+    }
+
+    public static AsyncMessageDispatcher encryptedMessageDispatcher(
+        byte[] privatePgpKey,
+        final String privatePgpKeyPassphrase
+    ) throws MessagingException {
+        if (empty(privatePgpKey) || empty(privatePgpKeyPassphrase)) {
```
adamvoncorswant: Minor: `allEmpty` is suitable here

izettle-messaging/src/main/java/com/izettle/messaging/handler/AsyncMessageDispatcher.java
xiaodong-izettle: Also, can we get rid of this `AsyncMessageDispatcher`? It's almost a duplication of `MessageDispatcher`, yes?

xiaodong-izettle:
Actually, if you want, you could just pass in an `ExecutorService` when creating `QueueProcessor`. Then you don't need to use `sendEmailAsync`, just use normal `sendEmail` in `email` service. The email sending will happen asynchronously, and we could still use the message deleting provided here... Does this suffice your use case?

xiaodong-izettle:
For the `ExecutorService` usage example, check `sun` or `merchant-report` repo...

powion:
That sounds like a better idea, I'll try it out. Thanks! Closing this for now.

PR #180, opened 2016-08-05
Cleanup: Remove CalendarTruncator

adamvoncorswant:
As it's not needed since java8 time API

ping @xiaodong-izettle


xiaodong-izettle:
Nice cleanup... 👍

PR #181, opened 2016-08-18
Accept native UUID string format in `createAlternative`

oskarwiksten:
* Accept UUIDs as strings in either base64 or native format in method `createAlternative`.
* Overload method to accept uuids in format `java.util.UUID`.


linnie:
Great @oskarwiksten , hope we soon all can align on using the native object. Things would be so much easier!! 👍 

FYI: @ssprang regarding having the planets only using UUID in it's native format and not b64

PR #182, opened 2016-09-09
Add izettle-dropwizard module

x1a0:
- [x] `CassandraSessionManaged`


izettle-dropwizard/src/main/java/com/izettle/dropwizard/cassandra/CassandraSessionManaged.java
```diff
@@ -0,0 +1,33 @@
...
+    }
+
+    @Override
+    public void start() {
+        // do nothing
+    }
+
+    @Override
+    public void stop() {
+        cluster.closeAsync();
```
jakanjakan: Closing the cluster means we don't need to close the session, right?

x1a0: Yep - http://docs.datastax.com/en/drivers/java/2.0/com/datastax/driver/core/Cluster.html#closeAsync--

jakanjakan: 👍🏻 !

PR #183, opened 2016-09-12
Register ParameterNamesModule so that @JsonCreator annotation works.

MrShish:
ParameterNamesModule enables the use of reflection when deserializing
json string. Instead of explicitly annotating with jsonparameter annoations,
you can simply use @JsonCreator annotation on the constructor.


trdarr:
👌

PR #184, opened 2016-09-16
TimeZoneId: Add fromString method

adamvoncorswant:
To easier map the standard string identifier, such as
'Stockholm/Europe', to the enum value.

ping @ssprang


izettle-java/src/test/java/com/izettle/java/TimeZoneIdTest.java
```diff
@@ -17,4 +19,11 @@ public void testGetZoneId() {
         Arrays.stream(TimeZoneId.values()).forEach(TimeZoneId::toZoneId);
     }
 
+    @Test
+    public void testFromString() {
+        for (TimeZoneId timeZoneId : TimeZoneId.values()) {
+            assertEquals(timeZoneId, TimeZoneId.fromString(timeZoneId.stringId));
```
ssprang: In the UserDTO we do `TimeZoneId().toZoneId().toString()` so shouldn't we be able to do that here too? (And keep the stringId private).

adamvoncorswant: You definitely have a point, but it turns out that both classes `TimeZone` and `ZoneId` takes freedom in changing the zone definition/name (while keeping the same offset). For example "HST" might be turned into "+10" which makes it hard to test in the above way, should we change as per your suggestion.

ssprang: I don't get it.. If we can't test using the same output that the UserDTO generates, then I won't be able to use this to parse the UserDTO in the product library..?
To clarify, we can't do?
```
assertEquals(timeZoneId, TimeZoneId.fromString(timeZoneId.toZoneId().toString()));
```

adamvoncorswant: No, we actually cannot. Reason is that the enum is based on the distinct number of available values from the `TimeZone` class. The `ZoneId` class merges some of these into the same value.

Anyway, I added all stringIds from `ZoneId` in the private map, and tweaked the comparison in the test. I think this is the way to go.

ssprang:
Ok, looks good, thanks for the assist @adamvoncorswant !

adamvoncorswant:
Thank you @ssprang !

PR #186, opened 2016-09-22
Add UUIDModule to message ObjectMapper.

Sarastro72:
In order to be able to deserialize both old style base64 UUIDs and native hex with dash format, I have added the UUIDModule to toolbox. I had to copy paste the class from API-library in order to avoid a circular dependency.


jakanjakan:
Nice, looking goood! 👏🏻

PR #187, opened 2016-09-27
Ensure schema update is applied to all

x1a0:


renatoathaydes:
We found out while running test migrations in the authentication server that the consistency level of the session used by the schema migrator does not matter because schema changes always propagate via gossip.

@alorlea can confirm?

I think we can close this PR.

PR #188, opened 2016-10-14
Make sure use resource path as directory

x1a0:
If the passed path parameter does not container `/` at the end, it will load files start with same name in resource dir.


Sarastro72:
Good stuff.

PR #189, opened 2016-11-08
Java 8:ify ValueChecks

trdarr:


adamvoncorswant:
Nice initiative! While I really like the beautiful syntax, I wonder if it's really worth it? I mean: the library has not changed from an external users perspective (and does not support more java8:ish syntax from the outside).

Also, these homegrown tests (apologies for the unscientific performance testing here):
```java
    @Test
    public void timeAllNull() {
        Object[] objs = new Object[]{null, new Object(), new Object(), new Object(), null};
        long totNanos = 0;
        for (int j = 0; j < 20; j++) {
            Instant start = Instant.now();
            for (int i = 0; i < 1000000; i++) {
                ValueChecks.allNull(objs);
            }
            Instant end = Instant.now();
            if(j < 10) {
                //ignore the first iterations to allow warmup
                continue;
            }
            totNanos += Duration.between(start, end).toNanos();
        }
        System.out.println("allNull ms = " + totNanos/1000000);
    }

    @Test
    public void timeWhiteSpace() {
        String test = "JUlinMBwW493sGRe0wFqDXik\t80J4StmiAs9ImFi YVWm25Jj9lH";
        long totNanos = 0;
        for (int j = 0; j < 20; j++) {
            Instant start = Instant.now();
            for (int i = 0; i < 1000000; i++) {
                ValueChecks.onlyWhitespace(test);
            }
            Instant end = Instant.now();
            if(j < 10) {
                //ignore the first iterations to allow warmup
                continue;
            }
            totNanos += Duration.between(start, end).toNanos();
        }
        System.out.println("onlyWhitespace ms = " + totNanos/1000000);
    }
```
When I ran them on `master`, I got the following output:
```
onlyWhitespace ms = 30
allNull ms = 28
```
When I ran them on this branch, I got the following:
```
onlyWhitespace ms = 316
allNull ms = 527
```

It's more than 10 times slower. Normally I wouldn't be bothered by such a performance penalty, but when it's in a util lib with small atomic, almost language like functions used in tons of places, I tend to do so.

What do you think: is is worth it?

trdarr:
When you put it that way, it's probably not worth it, no. The new code reads nicer but it's not like the old code is unreadable and no one is reading this code anyway. Closing!

adamvoncorswant:
Cool. Thanks for the effort anyway :cake:

PR #191, opened 2016-11-09
Fix NullPointerException in toString

trdarr:
Fix a `NullPointerException` that I introduced in 6e46fd2 (#190). The old code used `StringBuilder#append`, which uses `String.valueOf` to translate `null` elements to "null". The new code used `Object#toString`, which obviously throws a `NullPointerException` for `null` elements.

`Objects.toString`, like `StringBuilder#append`, uses `String.valueOf`.

### DIFF
--- a/izettle-java/src/main/java/com/izettle/java/CollectionUtils.java
+++ b/izettle-java/src/main/java/com/izettle/java/CollectionUtils.java
@@ -6,6 +6,7 @@ import java.util.Collection;
 import java.util.HashSet;
 import java.util.LinkedList;
 import java.util.List;
+import java.util.Objects;
 import java.util.Set;
 import java.util.SortedSet;
 import java.util.TreeSet;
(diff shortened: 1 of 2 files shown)


izettle-java/src/main/java/com/izettle/java/CollectionUtils.java
```diff
@@ -84,6 +85,6 @@ public static String toString(Collection<?> collection, String delimiter) {
             return null;
         }
 
-        return collection.stream().map(Object::toString).collect(Collectors.joining(delimiter));
+        return collection.stream().map(Objects::toString).collect(Collectors.joining(delimiter));
```
adamvoncorswant: Ah. Easy mistake!

trdarr: There wasn't a test for it...

PR #192, opened 2016-11-11
Check sendMessageBatch result, log if there are failures.

xiaodong-izettle:
How about something like this, @calle? Thanks...

### DIFF
--- a/izettle-messaging/src/main/java/com/izettle/messaging/QueueServiceSender.java
+++ b/izettle-messaging/src/main/java/com/izettle/messaging/QueueServiceSender.java
@@ -6,8 +6,10 @@ import static com.izettle.java.ValueChecks.empty;
 
 import com.amazonaws.AmazonServiceException;
 import com.amazonaws.services.sqs.AmazonSQS;
+import com.amazonaws.services.sqs.model.BatchResultErrorEntry;
 import com.amazonaws.services.sqs.model.SendMessageBatchRequest;
 import com.amazonaws.services.sqs.model.SendMessageBatchRequestEntry;
+import com.amazonaws.services.sqs.model.SendMessageBatchResult;
 import com.amazonaws.services.sqs.model.SendMessageRequest;
 import com.amazonaws.services.sqs.model.SendMessageResult;
 import com.fasterxml.jackson.core.JsonProcessingException;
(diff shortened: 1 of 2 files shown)


calle:
Yes, this is better than silently ignoring them. Maybe add the special case of sending only a single message and throw an exception to allow the caller to handle it better?

This is what is invoked when you invoke `post(T message, String eventName)` (and done when publishing card payments from the planet to the sun.

calle:
Also some tests are failing

xiaodong-izettle:
Hehe, sure, thanks, @calle. Will look into it... 👍

xiaodong-izettle:
Ping @calle, Oskar had already added `post(T message, String eventName)` method... 

Here I just add logging for batch operations...

izettle-messaging/src/main/java/com/izettle/messaging/QueueServiceSender.java
```diff
@@ -192,7 +201,34 @@ private String wrapInSNSMessage(
 
     private void sendMessageBatch(Collection<SendMessageBatchRequestEntry> messages) {
         for (Collection<SendMessageBatchRequestEntry> batch : partition(messages, MAX_BATCH_SIZE)) {
-            amazonSQS.sendMessageBatch(new SendMessageBatchRequest(queueUrl, new ArrayList<>(batch)));
+            final SendMessageBatchResult sendMessageBatchResult =
+                amazonSQS.sendMessageBatch(new SendMessageBatchRequest(queueUrl, new ArrayList<>(batch)));
+            final List<BatchResultErrorEntry> failed = sendMessageBatchResult.getFailed();
+            if (!failed.isEmpty()) {
```
calle: Any chance of a `NullPointerException` here? Or will `getFailed` always return a valid List?

xiaodong-izettle: Hehe, thanks, @calle... 

Based on the source code, it's always a `List`... :)

calle: The best type of argument! 😄

calle:
What do you say about throwing a `MessagingException` instead? A partial failure should probably be handled in the same way as a full failure right?

xiaodong-izettle:
Yeah, I was considering that... 

Then I also consider that should we keep the behaviour the same at first, to have an impression about the failure rate first? 

I don't know, what do you say? :-)

calle:
Let's start with logging since Oskar already updated the method planet uses (single message) to throw on errors and that was the important thing to do (to notice when these publications failed).

xiaodong-izettle:
Hehe, thanks, @calle. :-)

PR #193, opened 2016-11-14
Use `sqs.sendMessage` for single message instead of `postBatch`

oskarwiksten:
* Calls `sqs.sendMessage(SendMessageRequest)` when sending a single message through `QueueServiceSender`, instead of calling `sqs.sendMessageBatch` via `postBatch`
* Reason for changing is to have better handling of errors in case a single message fails.

Ping @xiaodong-izettle , how about this for solving the same issue that #192 intends to solve?


xiaodong-izettle:
Hehe, sure, thanks for fixing this... But IMHO, #192 is still needed, for batch posting messages... The client should at least have some information about whether the batch fails (or partially fails) or not, yes? :-)

PR #194, opened 2016-11-21
Bugfix: Specify `queueURL` when sending single sqs message

oskarwiksten:
* Specify the destination queue url when sending a single message through `MessagePublisher::post`.
* If not specifying the `queueUrl`, an exception of type `com.amazonaws.services.sqs.model.QueueDoesNotExistException` will be thrown when this is called.
* Bugfix for #193 .

Ping @xiaodong-izettle


xiaodong-izettle:
Sure, will merge after TC is OK... 👍

PR #195, opened 2016-11-28
Upgrade various libraries.

xiaodong-izettle:
F.Y.I @iZettle/horizons...


pom.xml
```diff
@@ -58,11 +58,12 @@
 
         <assertj-core.version>3.4.1</assertj-core.version>
         <astyanax.version>3.9.0</astyanax.version>
-        <aws.version>1.10.77</aws.version>
-        <bouncycastle.version>1.54</bouncycastle.version>
+        <aws.version>1.11.60</aws.version>
```
alorlea: I like this change :+1:

xiaodong-izettle:
Ping @iZettle/horizons, thanks...

izettle-cart/pom.xml
```diff
@@ -15,8 +15,8 @@
             <plugin>
                 <artifactId>maven-compiler-plugin</artifactId>
                 <configuration>
-                    <source>1.6</source>
-                    <target>1.6</target>
+                    <source>1.7</source>
+                    <target>1.7</target>
```
adamvoncorswant: I don't think this will work for our Android devs...

adamvoncorswant: Well, I think they're almost compatible with 1.7, but not all features (such as try-with-resources for example). So, while bumping source version wouldn't break compatibility itself, an unknowing backend coder could all of a sudden add code that won't work on android.

Thanks

fredrikbackstrom: Doesn't Android support J7 and sort of J8? https://developer.android.com/guide/platform/j8-jack.html

adamvoncorswant: Sort of, but not completely, and especially not API version 18 and below (which we still support)

xiaodong-izettle:
@adamvoncorswant, could you please merge this PR if it's OK? :-)

xiaodong-izettle:
@alorlea, since you are still in the office I assume?

alorlea:
Let me start the laptop

xiaodong-izettle:
Ah, sorry for the hassle... :-)

alorlea:
Merging

xiaodong-izettle:
Thanks, @alorlea. :)

adamvoncorswant:
Ah, thanks guys. Sorry for bad latency :)

PR #197, opened 2016-12-14
Java alb

danizettle:
adding a common package to register and de-register a dropwizard application in aws target group

ping @tbaylis , @fredrikbackstrom

### DIFF
--- a/izettle-java-alb/README.md
+++ b/izettle-java-alb/README.md
@@ -1,0 +1,7 @@
+# izettle-java-alb
+
+Module to use together with Dropwizard. Should be registered in the Application class 
+and will be called on start and stop of the application server, either during deployment or a 
+restart of the instance itself. 
+
+Authentication roles must be provided by IAM.
--- a/izettle-java-alb/README.md
+++ b/izettle-java-alb/README.md
@@ -3,5 +3,3 @@
 Module to use together with Dropwizard. Should be registered in the Application class 
 and will be called on start and stop of the application server, either during deployment or a 
 restart of the instance itself. 
-
-Authentication roles must be provided by IAM.
--- a/izettle-java-alb/src/main/java/com/izettle/alb/ALBAvailableInstancesCondition.java
+++ b/izettle-java-alb/src/main/java/com/izettle/alb/ALBAvailableInstancesCondition.java
@@ -9,7 +9,6 @@ import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.TimeUnit;
 import java.util.function.Consumer;
-import java.util.function.Supplier;
 
 /**
  * Precondition that check that enough instances are available in the ELB before the precondition is met.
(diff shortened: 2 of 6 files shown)


tbaylis:
Seems ok, but maybe some documentation describing how to use the package would be nice?

adamvoncorswant:
Also, nitpicking: please use spaces instead of tabs, and make sure each file has a newline at EOF.

izettle-java-alb/README.md
```diff
@@ -0,0 +1,7 @@
+# izettle-java-alb
+
+Module to use together with Dropwizard. Should be registered in the Application class 
+and will be called on start and stop of the application server, either during deployment or a 
+restart of the instance itself. 
+
+Authentication roles must be provided by IAM.
```
tbaylis: Don't we require a `AmazonElasticLoadBalancingClient`, not caring about where it gets its credentials from?

danizettle: True dat

izettle-java-alb/src/main/java/com/izettle/alb/Condition.java
```diff
@@ -0,0 +1,23 @@
...
+public interface Condition {
+
+    /**
+     * Await a precondition is met for a specified max duration.
+     * @param executorServiceSupplier Supplier of a executor service
+     * @param notificationsConsumer A consumer that can take string messages from the precondition
+     * @return
+     */
+    CompletableFuture<Boolean> check(
+        Supplier<ExecutorService> executorServiceSupplier,
```
xiaodong-izettle: Maybe just a plain `ExecutorService  executorService` is enough?

izettle-java-alb/src/main/java/com/izettle/alb/Condition.java
```diff
@@ -0,0 +1,23 @@
...
+
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutorService;
+import java.util.function.Consumer;
+import java.util.function.Supplier;
+
+/**
+ * Interface for classes that implement a condition check.
+ */
+public interface Condition {
```
xiaodong-izettle: Some nitpick, maybe add `@FunctionalInterface` for this interface, if it's intended to use `lambda` syntax... :)

izettle-java-alb/src/main/java/com/izettle/alb/ALBAvailableInstancesCondition.java
```diff
@@ -0,0 +1,99 @@
...
+
+    @Override
+    public CompletableFuture<Boolean> check(
+        Supplier<ExecutorService> executorServiceProvider,
+        Consumer<String> notificationsConsumer
+    ) {
+
+        return CompletableFuture.supplyAsync(
+            () -> {
+                while (true) {
```
danizettle: We will look at the ansible script while deploying so we will see if something is up. Only alternative is to fail the startup.

PR #198, opened 2016-12-14
TLVDecoder remove array alloc during recursion.

fredrikbackstrom:
This PR removes the new byte array allocation during recursive
calls. It reuses the ByteBuffer when the value of a tag is a new
TLV byte array. It keeps the array allocation when the value of a
tag is _just_ a "value", this should keep array allocations at a
minimum and my very non-educated guess is that the big arrays are the
ones where the value of a tag is a TLV, maybe @oskarwiksten have some 
more insight?

It also converts some of the recursive calls into sequential calls by looping, 
not expecting wonders out of this though.

### DIFF
--- a/izettle-emv/src/main/java/com/izettle/tlv/TLVDecoder.java
+++ b/izettle-emv/src/main/java/com/izettle/tlv/TLVDecoder.java
@@ -1,5 +1,6 @@
 package com.izettle.tlv;
 
+import java.nio.ByteBuffer;
 import java.util.ArrayList;
 import java.util.HashSet;
 import java.util.List;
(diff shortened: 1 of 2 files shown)


fredrikbackstrom:
I created a fix that is less of a risk https://github.com/iZettle/izettle-toolbox/pull/199 it does not remove as many assignments as this one but it removes the one that is most likely to be the big one.

lagex-iztl:
I like the simplicity of the change in https://github.com/iZettle/izettle-toolbox/pull/199

izettle-emv/src/main/java/com/izettle/tlv/TLVDecoder.java
```diff
@@ -32,43 +33,54 @@ public void addExpandTag(byte[] tag) throws TLVException {
...
+
+        final ByteBuffer wrap = ByteBuffer.wrap(in);
+
+        addAllTags(wrap, in.length, out);
+
         return out;
     }
 
-    private void helper(byte[] input, int offset, List<TLV> tags) throws TLVException {
+    private void addAllTags(ByteBuffer inpput, long maxLength, List<TLV> out) throws TLVException {
```
fredrikbackstrom: woops

PR #200, opened 2016-12-19
TLVDecoder: Prevent overflowing int when parsing TLV lengths

oskarwiksten:
* When checking a parsed TLV length, prevent input data from overflowing `Int.MAX_VALUE`.
* Prevents the following exception (as can be seen when running the unit-test `shouldDecodeTLVDataWithFourByteLengthOfAlmostIntMaxValue` added in this commit):

```
Stacktrace was: java.lang.OutOfMemoryError: Java heap space
	at com.izettle.tlv.TLVDecoder.helper(TLVDecoder.java:99)
	at com.izettle.tlv.TLVDecoder.decode(TLVDecoder.java:35)
	at com.izettle.tlv.TLVDecoderTest.shouldDecodeTLVDataWithFourByteLengthOfAlmostIntMaxValue(TLVDecoderTest.java:200)
```

Ping @linnie @xiaodong-izettle

### DIFF
--- a/izettle-emv/src/main/java/com/izettle/tlv/TLVDecoder.java
+++ b/izettle-emv/src/main/java/com/izettle/tlv/TLVDecoder.java
@@ -92,7 +92,7 @@ public class TLVDecoder {
 
         ++offset; // Now positioned at first data byte
 
-        if (offset + length > input.length) {
+        if (length > input.length - offset) {
             throw new TLVException("Tag " + Hex.toHexString(tag) + " exceeds data length");
         }
 
--- a/izettle-emv/src/test/java/com/izettle/tlv/TLVDecoderTest.java
+++ b/izettle-emv/src/test/java/com/izettle/tlv/TLVDecoderTest.java
@@ -1,5 +1,6 @@
 package com.izettle.tlv;
 
+import static org.hamcrest.CoreMatchers.startsWith;
 import static org.junit.Assert.assertEquals;
 
 import java.util.List;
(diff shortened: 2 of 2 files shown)


izettle-emv/src/main/java/com/izettle/tlv/TLVDecoder.java
```diff
@@ -92,7 +92,7 @@ private void helper(byte[] input, int offset, List<TLV> tags) throws TLVExceptio
 
         ++offset; // Now positioned at first data byte
 
-        if (offset + length > input.length) {
+        if (length > input.length - offset) {
```
xiaodong-izettle: 👍 👏

xiaodong-izettle:
Nice work 👍

fredrikbackstrom:
@oskarwiksten how does this relate to the other fixes? https://github.com/iZettle/izettle-toolbox/pull/198 and https://github.com/iZettle/izettle-toolbox/pull/199 ?

We might have different JVM:s that behave differently but neither shouldDecodeTLVDataWithFourByteLengthOfAlmostIntMaxValue nor shouldDecodeTLVDataWithFourByteLengthOfIntMaxValue gives a Java heap space OOM for me but a Requested array size exceeds VM limit OOM. I haven't digged into this myself but I tried fixing the Heap Space issue which at least on my machine this does not, are you running with special JVM settings to get a Heap Space issue and not Array size? Was the issue in prod a Array size or a Heap Space? 

for next time it would be nice if you would copy me into the issue given that you knew I had PR:s doing the same thing.

xiaodong-izettle:
@fredrikbackstrom, set something like `-Xmx256m -Xms256m`, the test will throw OOM... :)

fredrikbackstrom:
ok, thanks for the clarification @xiaodong-izettle

xiaodong-izettle:
NP, @fredrikbackstrom. I should have pinged you...  :)
Oskar probably run the test with Maven... :)

PR #206, opened 2017-01-03
Use default consistency level

xiaodong-izettle:
The default consistency level from the driver is good enough for us... 

Also adding support to set `localDc` when creating cluster object...

Ping @x1a0, thanks...

### DIFF
--- a/izettle-dropwizard/src/main/java/com/izettle/dropwizard/cassandra/CassandraSessionFactory.java
+++ b/izettle-dropwizard/src/main/java/com/izettle/dropwizard/cassandra/CassandraSessionFactory.java
@@ -1,10 +1,8 @@
 package com.izettle.dropwizard.cassandra;
 
 import com.datastax.driver.core.Cluster;
-import com.datastax.driver.core.ConsistencyLevel;
 import com.datastax.driver.core.HostDistance;
 import com.datastax.driver.core.PoolingOptions;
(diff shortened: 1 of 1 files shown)


xiaodong-izettle:
Ping @x1a0, thanks... :)

x1a0:
@xiaodong-izettle [Here](http://docs.datastax.com/en/drivers/java/3.0/com/datastax/driver/core/policies/DCAwareRoundRobinPolicy.html) says:
> If used with a single data center, this policy is equivalent to the RoundRobinPolicy, but its DC awareness incurs a slight overhead so the latter should be preferred to this policy in that case.

I think we only have single DC? Or are we preparing for multiDC here?

xiaodong-izettle:
Hehe, @x1a0, we are using this `DCAwareRoundRobinPolicy` now, yes? And it seems the document is right that the overhead is really slight... :)

And yes, we only have single DC now. And we have to use multiDC sometime, like when migrating the cluster into VPC... Hope I made myself clear.. :)

x1a0:
@xiaodong-izettle Yep - I was thinking if we only need single DC then changing to use `RoundRobinPolicy` instead of `DCAwareRoundRobinPolicy` as doc suggests.
Since multiDC is needed soon I think this is good.
Thanks - I am merging it.

xiaodong-izettle:
Thanks for merging, @x1a0. :)

PR #208, opened 2017-02-01
Serialize Instants in Rfc3339 format.

MrShish:
Before this Instants got serialized as 2001-12-23T02:05:06.123Z 
instead of 2001-12 -23T02:05:06.123+0000

### DIFF
--- a/izettle-messaging/src/main/java/com/izettle/messaging/serialization/InstantRFC3339Module.java
+++ b/izettle-messaging/src/main/java/com/izettle/messaging/serialization/InstantRFC3339Module.java
@@ -1,0 +1,63 @@
+package com.izettle.messaging.serialization;
+
+import com.fasterxml.jackson.core.JsonGenerator;
+import com.fasterxml.jackson.core.JsonParser;
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.*;
+import com.fasterxml.jackson.databind.module.SimpleModule;
+import java.io.IOException;
+import java.time.Instant;
+import java.time.ZoneId;
+import java.time.format.DateTimeFormatter;
+
+/**
+ * A simple Jackson module that will serialize and deserialize an Instant in accordance with RFC3339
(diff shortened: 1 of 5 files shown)


silverpilen:
Nice @MrShish Looks good.
@trdarr @martinfridh Any input or concerns?

trdarr:
From RFC 3339:
> Date and time formats cause a lot of confusion and interoperability problems on the Internet.

izettle-messaging/src/main/java/com/izettle/messaging/serialization/InstantRFC3339Module.java
```diff
@@ -0,0 +1,63 @@
...
+
+/**
+ * A simple Jackson module that will serialize and deserialize an Instant in accordance with RFC3339
+ * The default jackson module will use only the Zulu 'Z' zone indicator, which breaks compatibility with
+ * quite a few clients. This module ensures that both variants can be parsed, but that +0000 is always used in the
+ * serialized representation.
+ *
+ * Note: This module needs to be registered after other possible modules that might try to control `Instant`, such as
+ * the JaveTimeModule.
+ */
```
trdarr: This answers my question.

adamvoncorswant:
@MrShish : this looks like a copy of the `InstantRFC3339Module` in api-library. Any specific reason for not using that?

martinfridh:
😬  maybe time to merge/consolidate the different izettle utils repos to avoid situations like this.....

adamvoncorswant:
@martinfridh : this repo is public and for generic utilities, while the aforementioned is for izettle specific needs when building internal APIs. But, yeah, sometimes it's hard to know which one to put it in.

MrShish:
@adamvoncorswant , then i would have to introduce a pom dependency to a private repo... is that kosher?

adamvoncorswant:
Ah, No, then I'm fine with this until we create a way to provide your own mapper here

trdarr:
Another gem from RFC 3339:

> Human readability has proved to be a valuable feature of Internet protocols.

PR #210, opened 2017-02-03
Remove InstantRFC3339Module from messaging lib

MrShish:
And use the one we have in izettle-jackson


silverpilen:
Sweet!
Waiting for TC

PR #213, opened 2017-02-16
Instant serializer: Add test verifying presence of ms

adamvoncorswant:
Even if the time is in even seconds.

Seems to pass. I still don't know what caused the absence of ms in our systems during the last couple of days.

Anyway, I thought that the test might be good to have in place anyway... ping @xiaodong-izettle


xiaodong-izettle:
👍 
Nice one...

adamvoncorswant:
Thanks @xiaodong-izettle

PR #215, opened 2017-03-15
INT-290: Add UUIDParamConverterProvider

Sarastro72:
Add a ParamConverterProvider for UUIDS that supports both standard hex format as well as base64.

This relates to thhis PR in cashregister-service. https://github.com/iZettle/cash-register/pull/170/files

As soon as this is merged I can release and bump the version for toolbox and use it in cash-register-service.


Sarastro72:
> Do we need unit test for this?

At first I thought not, but sure! @x1a0 please rereview.

adamvoncorswant:
:+1:

PR #216, opened 2017-03-28
Added taxable value to ItemLine

psvanstrom:
Used for getting the taxable value of a line before any VAT has been applied.


adamvoncorswant:
:cake:

PR #217, opened 2017-03-28
Cart: Add test for taxable value

adamvoncorswant:
Also change return type of `ItemLine::getActualTaxableValue` from `Long`
to `long` as it's always going to return a valid value, and the
convention in this project is to then use native types.

ping @psvanstrom


psvanstrom:
Looks good! 🍰  ⭐️

adamvoncorswant:
Thanks @psvanstrom

PR #218, opened 2017-03-30
Dependency mgmt: Exclude 3.2 from cassandra version

adamvoncorswant:
Previous maven range was open-ended and made a new version break the build. This
change simply requires the allowed version to be strictly less than 3.2


adamvoncorswant:
Thanks @Sarastro72 !

PR #219, opened 2017-04-03
Adds support for getting the amount of VAT per unit of a line

olsson:
### Background
In the UK, we must, [according to law](https://www.gov.uk/vat-record-keeping/vat-invoices), list the price per item excluding VAT on invoices. The cart library today supports listing price per _line_ and price per _cart_ excluding VAT. But this is not enough for the UK. 

### What does this PR change?
This PR adds a field to the `ItemLine` class called `unitVat`. `CartUtils.buildItemLines` is modified to also put the VAT amount per unit in the line. I consider this to be the most backwards-compatible way to accomplish this.

### Could this be done in a better way?
Yes, absolutely! 
We could modify `Item` to add a default method:
```java
/**
 * Returns the VAT amount per item
 * @return
 */
default long getUnitVat() {
    return CartUtils.calculateVatFromAmount(getUnitPrice(), getVatPercentage());
}
```
But we can't do this right now, because it would break Android support.

### DIFF
--- a/izettle-cart/src/main/java/com/izettle/cart/CartUtils.java
+++ b/izettle-cart/src/main/java/com/izettle/cart/CartUtils.java
@@ -309,8 +309,9 @@ class CartUtils {
             } else {
                 effectivePrice = ItemUtils.getValue(item);
             }
-            final Long grossVat = calculateVatFromGrossAmount(ItemUtils.getGrossValue(item), item.getVatPercentage());
-            final Long effectiveVat = calculateVatFromGrossAmount(effectivePrice, item.getVatPercentage());
+            final Long grossVat = calculateVatFromAmount(ItemUtils.getGrossValue(item), item.getVatPercentage());
+            final Long effectiveVat = calculateVatFromAmount(effectivePrice, item.getVatPercentage());
+            final Long unitVat = calculateVatFromAmount(item.getUnitPrice(), item.getVatPercentage());
 
             final ItemLine<T, K> itemLine = new ItemLine<T, K>(
                 item,
@@ -318,7 +319,8 @@ class CartUtils {
                 grossVat,
                 effectivePrice,
                 effectiveVat,
-                ItemUtils.getDiscountValue(item)
+                ItemUtils.getDiscountValue(item),
+                unitVat
             );
 
             retList.add(itemLine);
@@ -326,7 +328,7 @@ class CartUtils {
         return retList;
     }
 
-    static Long calculateVatFromGrossAmount(final Long amountIncVat, final Float vatPercent) {
+    static Long calculateVatFromAmount(final Long amountIncVat, final Float vatPercent) {
         if (amountIncVat == null || vatPercent == null) {
             return null;
         }
--- a/izettle-cart/src/test/java/com/izettle/cart/ItemTest.java
+++ b/izettle-cart/src/test/java/com/izettle/cart/ItemTest.java
@@ -80,6 +80,7 @@ public class ItemTest {
         ItemLine<TestItem, TestDiscount> line2 = itemLines.get(1);
         assertEquals(400L, line2.getActualValue());
         assertEquals(320L, line2.getActualTaxableValue());
+        assertEquals(200L, (long) line2.getUnitVat());
     }
 
     @Test
--- a/izettle-cart/src/main/java/com/izettle/cart/ItemLine.java
+++ b/izettle-cart/src/main/java/com/izettle/cart/ItemLine.java
@@ -10,7 +10,7 @@ import java.io.Serializable;
  * @param <T> The type of the Item itself
  * @param <K> The type of the optional Discount that might be associated with the provided Item
  */
-public class ItemLine<T extends Item<T, K>, K extends Discount<K>> implements Serializable{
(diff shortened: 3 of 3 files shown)


izettle-cart/src/test/java/com/izettle/cart/ItemTest.java
```diff
@@ -80,6 +80,7 @@ public void testTaxableValueIsCorrect() {
         ItemLine<TestItem, TestDiscount> line2 = itemLines.get(1);
         assertEquals(400L, line2.getActualValue());
         assertEquals(320L, line2.getActualTaxableValue());
+        assertEquals(200L, (long) line2.getUnitVat());
```
martinfridh: outside scope of this PR really, though using assertJ would be nicer here, and wouldn't force casting any values - `assertThat(200L).isEqualTo(line2.getUnitVat());`

olsson:
@adamvoncorswant The relevant information is in the list of what makes a valid invoice:
> Price per item, excluding VAT

adamvoncorswant:
I believe that an item is the line rather than a physical product in these contexts. If that's the case, getting the amount without VAT is already available in the cart lib. Otherwise, it's getting pretty hard to explain the unit VAT amount if you sell minced meat for 1 öre per gram. Still if it's 1000 grams on that line and you have 25% VAT,  the total VAT amount on that line will be 2.5kr. Communicating a rounded VAT amount per unit will open up a can of worms with rounding errors/questions.

Thoughts on this?

olsson:
@adamvoncorswant UK (B2B) invoices present everything excluding VAT, except the total line. 

If a company with a VAT number sends an invoice to another company, we should not show any VAT per item or per line. The only places we show VAT is at the bottom: total VAT amount and total including VAT. We do not need to show the actual unit VAT amount, but instead the unit price excluding VAT.  This is what we interpret to be the "price per item excluding VAT."

I chose to implement this as `getUnitVat()` to follow the existing pattern of getting the amount of VAT, not the adjusted unit price.  The invoice service will use the Cart library to deduct the `unitVat` from the `unitPrice` when rendering the invoice.

izettle-cart/src/main/java/com/izettle/cart/CartUtils.java
```diff
@@ -309,24 +309,26 @@ static Double getDiscountPercentage(long grossValue, Long discountValue) {
...
+                unitVat
             );
 
             retList.add(itemLine);
         }
         return retList;
     }
 
-    static Long calculateVatFromGrossAmount(final Long amountIncVat, final Float vatPercent) {
+    static Long calculateVatFromAmount(final Long amountIncVat, final Float vatPercent) {
```
adamvoncorswant: What's the benefit of removing 'gross' here? It's a huge difference between calculating VAT from a net or a gross value and the word served as a helpful hint for the reader, don't you think?

olsson: Now that you've explained the naming, it makes sense. I was trying to make it clearer for me. I will change it back.

adamvoncorswant:
> If a company with a VAT number sends an invoice to another company, we should not show any VAT per item or per line. The only places we show VAT is at the bottom: total VAT amount and total including VAT. We do not need to show the actual unit VAT amount, but instead the unit price excluding VAT. This is what we interpret to be the "price per item excluding VAT."

> I chose to implement this as getUnitVat() to follow the existing pattern of getting the amount of VAT, not the adjusted unit price. The invoice service will use the Cart library to deduct the unitVat from the unitPrice when rendering the invoice.

I see... Well, the problem is that this is never going to add up. You will have rounding errors on each line that'll accumulate and make people (and the British tax authorities) scratch their heads. I honestly don't know how this should be solved, except for making it possible for customers to specify their price per unit *without VAT* as opposed to today where all prices are entered *including VAT*. This, however, seems extremely burdensome to change...

olsson:
@mrshish thoughts on this?

adamvoncorswant:
Just to clarify:
A merchant enters a unit price of 100kr with a VAT percentage of 21%. The unit price without VAT will then be ~82.644628099173kr. If we have to present this price on the invoice, we'll have to show 82.64kr per unit.
Let's say a customer buys 3 units of this product. That will, as we do it today, be calculated as 3*100=300kr, where 300*0.21/1.21 ~ 52.07kr VAT. If we use the stated unit price without VAT, we'll get: 3*82.64=247.92kr, where 267.87*0.12~52.06kr VAT.

This was just an example: I'm certain that there be much higher errors for other prices and VAT rates. The effect will also, of course, be accumulated if a higher number of items is bought.

olsson:
@adamvoncorswant After some discussion with @izwarholm and @MrShish, we hypothesize that a 0.5p-1p discrepancy in VAT on these invoices is an acceptable error, and from @izwarholm's research, might actually be ignored by UK tax authorities. We will proceed this route, but leave the logic outside of the cart library, and isolate it for invoices only.

MrShish:
Suma sumarum: the total sum and total sum without vat is still valid, and what actually is needed to do a correct tax deduction..
The rounding errors is ok as long as we move it outside of the mathematical realm and treat it as a presentational aid.

izwarholm:
Some additional discussion (wall of text) https://www.taxation.co.uk/Articles/2009/01/14/239261/fraction-frustration

PR #221, opened 2017-04-07
TLV Decoder: Fail early on malformed TLV tags

oskarwiksten:
* Throw explicit `TLVException` when reading the tag would otherwise exceed the buffer length.
* Prevents the following exception (as seen in the added unit test added in this commit):
```
java.lang.ArrayIndexOutOfBoundsException: 4
	at com.izettle.tlv.TLVDecoder.helper(TLVDecoder.java:60)
	at com.izettle.tlv.TLVDecoder.decode(TLVDecoder.java:35)
	at com.izettle.tlv.TLVDecoderTest.shouldThrowExceptionOnMalformedTag(TLVDecoderTest.java:230)
```
* Reason for adding is to safeguard callers that only catch `TLVException` (and not `RuntimeException`s such as `ArrayIndexOutOfBoundsException`), so that callers also handle such malformed input correctly.

Ping @linnie

### DIFF
--- a/izettle-emv/src/main/java/com/izettle/tlv/TLVDecoder.java
+++ b/izettle-emv/src/main/java/com/izettle/tlv/TLVDecoder.java
@@ -57,6 +57,9 @@ public class TLVDecoder {
              * tag. Subsequent tag bytes have 0x80 bit set.
              */
             do {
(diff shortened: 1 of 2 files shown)


linnie:
Nice - I like this! 🌼

PR #223, opened 2017-04-26
[DON'T MERGE YET] Upgraded aws lib & added support for container roles

freddd:


freddd:
Haven't found a project using this yet. Email, Image, AuthZ/N is not. Any ideas of where I could test it? @nzroller

PR #227, opened 2017-06-13
Moves AdminDirectAccessFilter to izettle-filters

nzroller:
Moves the admin filter to a new artifact `izettle-filters` with no
dependencies except for the servlet API.

The artifact `izettle-dropwizard` will be deprecated since it is not
widely used internally at iZettle (with the exception of the admin
filter).


rickard-von-essen-iz:
👍

PR #228, opened 2017-06-16
TimeZoneId: Update with current zones

adamvoncorswant:
And remove a few outdated ones.

ping @olsson

### DIFF
--- a/izettle-java/src/main/java/com/izettle/java/TimeZoneId.java
+++ b/izettle-java/src/main/java/com/izettle/java/TimeZoneId.java
@@ -23,7 +23,6 @@ public enum TimeZoneId {
     AMERICA_ADAK("America/Adak"),
     AMERICA_ATKA("America/Atka"),
     ETC_GMT_PLUS10("Etc/GMT+10"),
-    HST("HST"),
     PACIFIC_FAKAOFO("Pacific/Fakaofo"),
     PACIFIC_HONOLULU("Pacific/Honolulu"),
     PACIFIC_JOHNSTON("Pacific/Johnston"),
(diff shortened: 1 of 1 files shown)


izettle-java/src/main/java/com/izettle/java/TimeZoneId.java
```diff
@@ -653,6 +628,10 @@
         this.stringId = stringId;
     }
 
+    public String getStringId() {
+        return stringId;
+    }
+
     public TimeZone getTimeZone() {
         return TimeZone.getTimeZone(this.stringId);
```
olsson: Why not fix this too? This was the cause of the bug in Invoice. I think we should just return ZoneId using `ZoneId.of(this.stringId)`. I don't think we should mix `java.util.Date` classes with `java.time` classes unless its specifically for conversion logic. If we want to keep this, it should be made abundantly clear that this `toZoneId()` does its conversion via `java.util.TimeZone`

olsson: Even if we do keep the existing `toZoneId()` method, it's not clear to users of this enum that they should instead use `ZoneIf.of(TimeZoneId.getStringId())` to avoid potential issues with `java.util.TimeZone`

adamvoncorswant: Agreed! I now changed so that both methods uses `ZoneId` on the way out instead of `TimeZone`

adamvoncorswant:
I think so. Otherwise, I'll ping @nzroller after this is merged, but I would need a merge first :)

adamvoncorswant:
Nice. Thank you @olsson !

PR #229, opened 2017-07-11
InstantParam and a new DateTimeFormatter for RFC3339 formats

izrobin:
So the need to have a way of receiving an `Instant` either via `@QueryParam` or `@PathParam` increases, so I present to you: `InstantParam`.

It's basically a wrapper around an Instant, which assumes the query string or path variable, is formatted into RFC3339 date time format.
Also added a convenience method to `DateFormatCreator` for creating a new jdk 8 `DateTimeFormatter` (instead of the old `DateFormat`), so now you can parse & format `TemporalAccessor`s!

### DIFF
--- a/izettle-jackson/src/main/java/com/izettle/jackson/paramconverter/InstantParam.java
+++ b/izettle-jackson/src/main/java/com/izettle/jackson/paramconverter/InstantParam.java
@@ -1,0 +1,39 @@
+package com.izettle.jackson.paramconverter;
+
+import static com.izettle.java.DateTimeFormatters.RFC_3339_INSTANT;
+
+import java.time.Instant;
+import java.time.format.DateTimeParseException;
+import javax.ws.rs.BadRequestException;
+
+/**
+ * Used as a <b>@QueryParam</b> to wrap an Instant. The URI string representation should be in RFC3339 format
+ * Can also be used with to wrap an Instant with usage in <b>@PathParam</b>
+ */
+
+public class InstantParam {
+
+    private final Instant instant;
+
+    public InstantParam(final Instant instant) {
+        this.instant = instant;
+    }
+
+    public static InstantParam valueOf(final String rfc3339DateString) throws BadRequestException {
+        try {
+            final Instant instant = RFC_3339_INSTANT.parse(rfc3339DateString, Instant::from);
+            return new InstantParam(instant);
+        } catch (final DateTimeParseException e) {
+            throw new BadRequestException(e);
+        }
+    }
+
+    public Instant getInstant() {
+        return instant;
+    }
+
+    @Override
+    public String toString() {
+        return RFC_3339_INSTANT.format(instant);
+    }
+}
--- a/izettle-jackson/src/test/java/com/izettle/jackson/paramconverter/InstantParamTest.java
+++ b/izettle-jackson/src/test/java/com/izettle/jackson/paramconverter/InstantParamTest.java
@@ -1,0 +1,52 @@
+package com.izettle.jackson.paramconverter;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
+
+import java.lang.reflect.Method;
+import java.lang.reflect.Modifier;
+import java.time.Instant;
+import java.time.format.DateTimeParseException;
+import javax.ws.rs.BadRequestException;
+import org.junit.Test;
+
+public class InstantParamTest {
+
+    private Instant myInstant = Instant.ofEpochMilli(1387920896123L); // 24 Dec 2013, 22:34:56.123 (UTC+1)
+
+    @Test
+    public void shouldBeAbleToSerializeAnInstant() throws Exception {
+        final String result = new InstantParam(myInstant).toString();
+        assertEquals("2013-12-24T21:34:56.123+0000", result);
+    }
+
+    @Test
+    public void shouldHavePublicStaticValueOfMetod() throws Exception {
+        final Method valueOfMethod = InstantParam.class.getMethod("valueOf", String.class);
+        assertNotNull(valueOfMethod);
+        assertTrue(Modifier.isPublic(valueOfMethod.getModifiers()));
+        assertTrue(Modifier.isStatic(valueOfMethod.getModifiers()));
+        assertEquals(InstantParam.class, valueOfMethod.getReturnType());
+    }
+
+    @Test
+    public void shouldBeAbleToDeserializeIncomingRequestParameter() throws Exception {
+        final String requestParamValue = "2013-12-24T21:34:56.123+0000";
+        final InstantParam result = InstantParam.valueOf(requestParamValue);
+
+        assertEquals(myInstant, result.getInstant());
+    }
+
+    @Test
+    public void shouldThrowWhenDeserializeFaultyRequestParameter() throws Exception {
+        final String requestParamValue = "99-12-24T21:34:56.123+0000";
+        try {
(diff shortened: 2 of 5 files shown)


izettle-jackson/src/test/java/com/izettle/jackson/paramconverter/InstantParamTest.java
```diff
@@ -0,0 +1,38 @@
...
+    private Instant myInstant = Instant.ofEpochMilli(1387920896123L); // 24 Dec 2013, 22:34:56.123 (UTC+1)
+
+    @Test
+    public void shouldBeAbleToSerializeAnInstant() throws Exception {
+        final String result = new InstantParam(myInstant).toString();
+        assertEquals("2013-12-24T21:34:56.123+0000", result);
+    }
+
+    @Test
+    public void shouldHavePublicStaticValueOfMetod() throws Exception {
```
goncalosmo: :+1:

izettle-jackson/src/test/java/com/izettle/jackson/paramconverter/InstantParamTest.java
```diff
@@ -0,0 +1,38 @@
...
+        assertEquals(InstantParam.class, valueOfMethod.getReturnType());
+    }
+
+    @Test
+    public void shouldBeAbleToDeserializeIncomingRequestParameter() throws Exception {
+        final String requestParamValue = "2013-12-24T21:34:56.123+0000";
+        final InstantParam result = InstantParam.valueOf(requestParamValue);
+
+        assertEquals(myInstant, result.getInstant());
+    }
```
goncalosmo: What about another test that makes sure it throws an exception (even BadRequest, maybe) when the format is not expected? Just to make sure we don't get `null` into the resource.

izrobin: You'll get an `java.time.format.DateTimeParseException`, but perhaps we should catch it inside `InstantParam::valueOf` and convert it to a `BadRequestException` since that is more fitting for where it's going to be used. What do you think?

goncalosmo: My thought exactly. Not sure how Jersey will react to that, though.

izrobin: Did some testing locally in docker, throwing `BadRequestException` gives you the expected 400 Bad Request response, which is good. `IllegalArgumentException` gives you 404, which is weird. I think we're going to throw `BadRequestException`

goncalosmo: Great!

izettle-java/src/main/java/com/izettle/java/DateFormatCreator.java
```diff
@@ -43,6 +50,16 @@ public static DateFormat createRFC3339Formatter() {
     }
 
     /**
+     * Create a JDK 8 java.time.DateTimeFormatter commonly used when communicating with services over the internet.
+     * Has the ability to parse & format TemporalAccessors, such as java.time.Instant
+     * Example output: "2013-12-24T21:34:56.123+0000".
+     * @return the newly created formatter
+     */
+    public static DateTimeFormatter createRFC3339DateTimeFormatter() {
+        return createJdk8Formatter("yyyy-MM-dd'T'HH:mm:ss.SSSZ", TimeZoneId.UTC);
```
adamvoncorswant: The thing with these `createXXXformatter`-methods in this (old) class is that they were all about `java.util.Date` stuff. These classes weren't thread-safe and needed a new instance for each calling thread. Now though, with the `java.time`-classes in java8, they are all thread-safe and need not be initialized like this. One single static final instance of the `DateTimeFormatter` is enough. Wanna declare it as such instead?

adamvoncorswant: Much like for example `DateTimeFormatter.ISO_INSTANT`, but with our preferred format instead

adamvoncorswant: What about just putting it in another class, `DateTimeFormatters.RFC_3339_INSTANT`? Then we can even have as an ambition to completely remove this class (down the road when we have no usages of `java.util.Date` anymore.

izettle-java/src/main/java/com/izettle/java/DateFormatCreator.java
```diff
@@ -23,6 +24,12 @@ public static DateFormat createFormatter(String pattern, Locale locale, TimeZone
         return df;
     }
 
+    public static DateTimeFormatter createJdk8Formatter(final String pattern, final TimeZoneId timeZoneId) {
```
adamvoncorswant: Nitpick: also I'm not too fond of putting the version of the jdk in method names like this. Hopefully it will be equally useful for jdk9, 10 and 11 :)

izettle-java/src/main/java/com/izettle/java/DateFormatCreator.java
izrobin: I like it! Phase out the old stuff. Will fix

adamvoncorswant:
Looking really good! Now my only question is whether we should parse in both `InstantParam` and `InstantRfc339Module` with the same method? Right now they are parsing using different formatters...

izrobin:
Absolutely, but... I feel confused when reading the javadoc for `InstantRfc339Module`, beacuse it's backwards. Shouldn't Zulu-indicator be used when serializing? And deserializing should be able to do both Zulu and +0000? @adamvoncorswant did I get it right?

adamvoncorswant:
Admittedly I'm also confused. 😁 I'll swing by you when I come into the office!

izrobin:
Ok some clearifactions added, and I'm keeping the `InstantRfc339Module` formatters, as they are very specific to the deserialization for jackson (only). We should use the new `DateTimeFormatters. RFC_3339_INSTANT` for everything else

izettle-jackson/src/main/java/com/izettle/jackson/paramconverter/InstantParam.java
```diff
@@ -0,0 +1,40 @@
...
+
+public class InstantParam {
+
+    private final Instant instant;
+
+    public InstantParam(final Instant instant) {
+        this.instant = instant;
+    }
+
+    @SuppressWarnings("unused") //Used by reflection via @QueryParam / @PathParam
```
adamvoncorswant: Is this annotation really necessary for a public static method? No IDE in to my knowledge can say with correctness that no one is using such a method...

adamvoncorswant:
Just one last comment, otherwise I'm happy. Please let me know when you want this merged!

izettle-jackson/src/main/java/com/izettle/jackson/paramconverter/InstantParam.java
izrobin: IntelliJ complained that it was not being used (Before I had the unit tests, which of course uses the method) But I'm up for deleting it. We have unit tests that requires the method to exist :D

izrobin: I will delete it and squash all commits into one

izrobin: Merge away! @adamvoncorswant

adamvoncorswant:
:+1:

izrobin:
Can I do a release of toolbox? :) @adamvoncorswant

adamvoncorswant:
Yes, please do so!

PR #230, opened 2017-07-13
InstantParam: Improved parsing capabilities

adamvoncorswant:
So that it's identical to the `InstantRFC3339Module` in the sense that
it can also:
* parse strings using the Zulu indicator
* parse strings with none or parts of fractional seconds
* use 'uuuu' (year) instead of 'yyyy' (year-of-era) for serializing the
  year

Sorry for modifying your code from yesterday @izrobin but I didn't really get your underlying reason for not making the parsing more forgiving, as in the `InstantRFC3339Module`. Making the two of them behave identical should be a goal, no?

Anyway, please have a look @izrobin

### DIFF
--- a/izettle-jackson/src/main/java/com/izettle/jackson/paramconverter/InstantParam.java
+++ b/izettle-jackson/src/main/java/com/izettle/jackson/paramconverter/InstantParam.java
@@ -1,5 +1,7 @@
 package com.izettle.jackson.paramconverter;
 
+import static com.izettle.java.DateTimeFormatters.INSTANT_WITH_NO_MILLIS_FALLBACK;
+import static com.izettle.java.DateTimeFormatters.INSTANT_WITH_ZULU_OR_OFFSET;
 import static com.izettle.java.DateTimeFormatters.RFC_3339_INSTANT;
 
 import java.time.Instant;
@@ -21,10 +23,15 @@ public class InstantParam {
 
     public static InstantParam valueOf(final String rfc3339DateString) throws BadRequestException {
         try {
-            final Instant instant = RFC_3339_INSTANT.parse(rfc3339DateString, Instant::from);
+            final Instant instant = INSTANT_WITH_ZULU_OR_OFFSET.parse(rfc3339DateString, Instant::from);
             return new InstantParam(instant);
         } catch (final DateTimeParseException e) {
-            throw new BadRequestException(e);
+            try {
+                final Instant instant =  INSTANT_WITH_NO_MILLIS_FALLBACK.parse(rfc3339DateString, Instant::from);
+                return new InstantParam(instant);
+            } catch (final DateTimeParseException e2) {
+                throw new BadRequestException(e2);
+            }
         }
     }
 
--- a/izettle-jackson/src/test/java/com/izettle/jackson/paramconverter/InstantParamTest.java
+++ b/izettle-jackson/src/test/java/com/izettle/jackson/paramconverter/InstantParamTest.java
@@ -49,4 +49,36 @@ public class InstantParamTest {
             assertTrue(ex.getCause() instanceof DateTimeParseException);
         }
     }
+
+    @Test
+    public void shouldBeAbleToDeserializeIncomingRequestParameterWithZuluIndicator() throws Exception {
+        final String requestParamValue = "2013-12-24T21:34:56.123Z";
+        final InstantParam result = InstantParam.valueOf(requestParamValue);
+        assertEquals(myInstant, result.getInstant());
+    }
+
+    @Test
+    public void shouldBeAbleToDeserializeIncomingRequestParameterWithoutMillis() throws Exception {
+        final String requestParamValue = "2013-12-24T21:34:56+0000";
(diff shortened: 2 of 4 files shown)


izettle-jackson/src/test/java/com/izettle/jackson/paramconverter/InstantParamTest.java
```diff
@@ -49,4 +49,36 @@ public void shouldThrowWhenDeserializeFaultyRequestParameter() throws Exception
...
+        final Instant expected = Instant.parse("2013-12-24T21:34:56.21Z");
+        assertEquals(expected, result.getInstant());
+    }
+
+    @Test
+    public void shouldBeAbleToDeserializeIncomingRequestParameterWithZuluIndicatorWithoutMillis() throws Exception {
+        final String requestParamValue = "2013-12-24T21:34:56Z";
+        final InstantParam result = InstantParam.valueOf(requestParamValue);
+        final Instant expected = Instant.parse("2013-12-24T21:34:56Z");
+        assertEquals(expected, result.getInstant());
```
adamvoncorswant: Note: all the four tests above fails without the new changes.

izettle-java/src/main/java/com/izettle/java/DateTimeFormatters.java
```diff
@@ -4,7 +4,7 @@
 
 public class DateTimeFormatters {
 
-    private static final String RFC_3339_OFFSET_FORMAT = "yyyy-MM-dd'T'HH:mm:ss.SSSZ";
+    private static final String RFC_3339_OFFSET_FORMAT = "uuuu-MM-dd'T'HH:mm:ss.SSSZ";
```
adamvoncorswant: This was interesting. I didn't know that `uuuu` is the way to go. And all these years I have been doing it wrong. You learn every day! :D

izrobin: Nice, that must be new. Year-of-era doesn't sound so useful for us :)

izettle-jackson/src/main/java/com/izettle/jackson/paramconverter/InstantParam.java
```diff
@@ -21,10 +23,15 @@ public InstantParam(final Instant instant) {
 
     public static InstantParam valueOf(final String rfc3339DateString) throws BadRequestException {
         try {
-            final Instant instant = RFC_3339_INSTANT.parse(rfc3339DateString, Instant::from);
+            final Instant instant = INSTANT_WITH_ZULU_OR_OFFSET.parse(rfc3339DateString, Instant::from);
             return new InstantParam(instant);
         } catch (final DateTimeParseException e) {
-            throw new BadRequestException(e);
+            try {
+                final Instant instant =  INSTANT_WITH_NO_MILLIS_FALLBACK.parse(rfc3339DateString, Instant::from);
```
izrobin: But why would we want to have fallback logic in this new util class? It' only "serialized" in one way via `toString()`, and if you use this class as a `queryParam("something", new InstantParam(instant))`, it will be consistent all over.

izrobin: Isn't it best to set a standard right away? :P "+0000" offset is the way to go, both ways

adamvoncorswant: While I'd normally agree that it's a good thing to be stringent and only allow one format, I'd for this case argue that:
* The format we're using is loosing ground to the ISO standard one (see for example `Instant::toString`, and we shouldn't force clients (including 3:rd parties) to adopt our *exact* preferences.
* Some clients have other libraries backing their serializers, resulting in them discarding trailing millisecond zeroes (I believe that our Android clients already does this).
* In this way, we will always produce consistent output (RFC3339), while still allowing for clients to use the ISO standard.
* It's valuable to be consistent between query parameters and body contents, so having the `InstantParam` and the `InstantRFC3339Module` behave identically is something to strive for. Fewer surprises.

izrobin: Ok I see your point there :) I thought sending `Instant`s in the querystring was something fairly "new", but I guess it has been handled differently across the repos.
👍  Thanks for an explanation :)

izettle-java/src/main/java/com/izettle/java/DateTimeFormatters.java
izrobin: I'm thinking, scrap the static string, inline it at line 17 instead, like the other ones :)

adamvoncorswant: Ok! BRB

adamvoncorswant: Inlined the format string, also updated the class comment. Rebased and force pushed.

adamvoncorswant:
Yes please :)

adamvoncorswant:
Thanks @izrobin : I'll perform a maven release

PR #232, opened 2017-08-21
Instants: let InstantRFC3339Module be able to parse epoch longs.

MrShish:
The regular Date module for jackson can parse epoch longs as a Date.
Some entities has the epoch longs as the seriliazed version. These entities wont be able to use Instants transparently if
support for parsing epoch timestamps is not added.

### DIFF
--- a/izettle-jackson/src/main/java/com/izettle/jackson/module/InstantRFC3339Module.java
+++ b/izettle-jackson/src/main/java/com/izettle/jackson/module/InstantRFC3339Module.java
@@ -57,10 +57,20 @@ public class InstantRFC3339Module extends SimpleModule {
             final DeserializationContext context
         ) throws IOException, JsonProcessingException {
             final String value = jp.readValueAs(String.class);
+
             try {
                 return Instant.from(INSTANT_WITH_ZULU_OR_OFFSET.parse(value));
             } catch (DateTimeParseException e) {
-                return Instant.from(INSTANT_WITH_NO_MILLIS_FALLBACK.parse(value));
+                try {
+                    return Instant.from(INSTANT_WITH_NO_MILLIS_FALLBACK.parse(value));
+                } catch (DateTimeParseException e2) {
+                    try {
+                        long timeStamp = Long.parseLong(value);
+                        return Instant.ofEpochMilli(timeStamp);
+                    } catch (NumberFormatException e3) {
+                        throw e;
+                    }
+                }
             }
         }
     }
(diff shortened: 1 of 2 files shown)


MrShish:
sorry for the fuggly nested `try` `catch` but didn't wanna use regexp for this.
And should we do it like this? im open for suggestion

izettle-jackson/src/main/java/com/izettle/jackson/module/InstantRFC3339Module.java
```diff
@@ -57,10 +57,20 @@ public Instant deserialize(
...
             } catch (DateTimeParseException e) {
-                return Instant.from(INSTANT_WITH_NO_MILLIS_FALLBACK.parse(value));
+                try {
+                    return Instant.from(INSTANT_WITH_NO_MILLIS_FALLBACK.parse(value));
+                } catch (DateTimeParseException e2) {
+                    try {
+                        long timeStamp = Long.parseLong(value);
+                        return Instant.ofEpochMilli(timeStamp);
+                    } catch (NumberFormatException e3) {
+                        throw e;
```
adamvoncorswant: So now all failures will result in a NumberFormatException? I think that's weird, and that it's important that we, in the end, always throw a DateTimeParseException for failures...

MrShish: no, the `throw e;` throws the first exception, which is a `DateTimeParseException`

adamvoncorswant: Ah, silly me. You're right of course!

adamvoncorswant: BTW: It's too bad that we don't have `Optional::or` until java9. I hate this arrowhead `try`-`catch` kind of notation. Only thing I could come up with, given the current language was
```java
        @Override
        public Instant deserialize(
            final JsonParser jp,
            final DeserializationContext context
        ) throws IOException, JsonProcessingException {
            try {
                return Instant.ofEpochMilli(Long.parseLong(value));
            } catch (Exception e) {
                //intentionally ignoring
            }
            try {
                return Instant.from(INSTANT_WITH_NO_MILLIS_FALLBACK.parse(value));
            } catch (Exception e) {
                //intentionally ignoring
            }
            return Instant.from(INSTANT_WITH_ZULU_OR_OFFSET.parse(value));
        }

```
What do you think: is it prettier or not?

MrShish: tbh, I think this will be much slower, as it will probably throw NumberFormatException 99% of the time, and not as a last resort. 

And regarding how it looks, Its almost as equally bad 😜

(actually, i would like to update the bad rows, but cassandra is such a mess to work with)

adamvoncorswant: Aight! I think I agree :D

MrShish:
thanks chief!

PR #234, opened 2017-10-05
Add AsyncPublisherService

minisu:
In order to send messages in a non-blocking manner.


minisu:
Something like this should do the trick, wouldn't it @qinjin-izettle ?

minisu:
Hmm... I wonder if it's worth making an `AbstractMessagingPublisher` and then extend it with `MessagingPublisher` and `AsyncMessagingPublisher`.

minisu:
Pinging @karlgrund since I know he has battled this at least once (working around it using `CompletableFuture.runAsync` with custom thread pool).

xiaodong-izettle:
Wait a bit please. 😄

minisu:
@xiaodong-izettle No problem, I set it to `[DON'T MERGE]` :-)

izettle-messaging/src/main/java/com/izettle/messaging/AsyncPublisherService.java
```diff
@@ -0,0 +1,121 @@
...
+                    public void onError(Exception e) {
+                        LOG.warn("Failed to publish message " + eventName, e);
+                    }
+
+                    @Override
+                    public void onSuccess(PublishRequest request, PublishResult publishResult) {
+                        // Do nothing
+                    }
+                }
+            );
```
minisu: This is the only changed statement from `PublisherService`.

izettle-messaging/src/main/java/com/izettle/messaging/AsyncPublisherService.java
```diff
@@ -0,0 +1,121 @@
...
+public class AsyncPublisherService implements MessagePublisher {
+
+    private static final Logger LOG = LoggerFactory.getLogger(AsyncPublisherService.class);
+
+    private final String topicArn;
+    private final AmazonSNSAsync amazonSNS;
+    private final MessageSerializer messageSerializer;
+
+    public static MessagePublisher nonEncryptedPublisherService(AmazonSNSAsync client, final String topicArn) {
+        return AsyncPublisherService.nonEncryptedPublisherService(client, topicArn, new DefaultMessageSerializer());
```
qinjin-izettle: No need to state`AsyncPublisherService`, just `nonEncryptedPublisherService(client, topicArn, new DefaultMessageSerializer())` would be fine.

izettle-messaging/src/main/java/com/izettle/messaging/AsyncPublisherService.java
```diff
@@ -0,0 +1,121 @@
...
+        try {
+            String jsonBody = messageSerializer.serialize(message);
+            String encryptedBody = messageSerializer.encrypt(jsonBody);
+            PublishRequest publishRequest = new PublishRequest(topicArn, encryptedBody, eventName);
+            amazonSNS.publishAsync(
+                publishRequest,
+                new AsyncHandler<PublishRequest, PublishResult>() {
+                    @Override
+                    public void onError(Exception e) {
+                        LOG.warn("Failed to publish message " + eventName, e);
```
qinjin-izettle: How about `LOG.warn("Failed to publish message {} for {}", eventName, e.getMessage());` Do we really need to print stack trace?

izettle-messaging/src/main/java/com/izettle/messaging/AsyncPublisherService.java
```diff
@@ -0,0 +1,121 @@
...
+     * @throws MessagingException Failed to post message.
+     */
+    public <M> void post(M message, String eventName) throws MessagingException {
+
+        if (empty(eventName)) {
+            throw new MessagingException("Cannot publish message with empty eventName!");
+        }
+        try {
+            String jsonBody = messageSerializer.serialize(message);
+            String encryptedBody = messageSerializer.encrypt(jsonBody);
```
qinjin-izettle: I guess this will be time consuming, so warp the post method into `CompletableFuture` might be a better idea.

minisu: Encryption is very fast. For example, you can encrypt AES in Java 8 at 3 MB per _millisecond_ 💨

izettle-messaging/src/main/java/com/izettle/messaging/AsyncPublisherService.java
minisu: You're right. The reason it's this way is because it's a pure copy-paste of `PublisherService` (except for the things that had to change to make it async).

minisu: I can fix this anyway though.

izettle-messaging/src/main/java/com/izettle/messaging/AsyncPublisherService.java
minisu: I'm just copying the error message that was used in `PublisherService`.

I prefer to change as little as possible, in order to:

1. Make it as easy as possible for users to change from `PublisherService` to `AsyncPublisherService` (no need to change Splunk alerts, etc.).
2. Minimize the risk of introducing bugs.
3. Minimize the risk of breaking requirements, since I don't know the requirements. The only thing I know is that `PublisherService` is working for all teams' needs (except that it's blocking).

minisu: I realize that I should have been clearer with my intentions in the PR description.

qinjin-izettle: Ok, that sounds a good explanation, thanks!

minisu:
@xiaodong-izettle anything in particular? :-)

xiaodong-izettle:
@minisu, actually a lot. :)

1. This post message operation, if we want an async version, we definitely need to keep the sync version. Because, we do have cases that sync operation is a necessity, as someone already mentioned in a Slack channel. Or if you want examples, check `AmazonSNSAsyncClient`, or Cassandra native driver. We cannot ask the API client to create two publisher instances, one for sync operations, the other for async ones. 

2. We definitely need to consider failure cases when doing async operations, pure logging the failure is not enough sometimes. Not all scenario fits fire-and-forget.

3. This publish message operation is not slow (relatively of course) by itself, it's just a Http call to the SNS endpoint. 

4. This is a public repo, we should not push a not mature solution to the public. 

But great initiative, @minisu, 👍. I'll definitely look into this. 😄

minisu:
> 1. We cannot ask the API client to create two publisher instances, one for sync operations, the other for async ones.

It's not obvious to me why we cannot ask API clients to do so. But I'm not against it either. Feel free to modify this PR.

> 2. We definitely need to consider failure cases when doing async operations, pure logging the failure is not enough sometimes. Not all scenario fits fire-and-forget.

This PR is not trying to solve all scenarios, and I don't think it has to. The most common scenario, in both my current and last team, has been to send analytics messages only intended for our Growth department. These events are far from critical and a `LOG.warn` should be enough.

Luckily, the added method (` public <M> void post(M message, String eventName)`) is future-compatible, in the sense that you can either modify it to return a `CompletableFuture` or add an overloaded method that also takes a callback as parameter.


> 3. This publish message operation is not slow [...]

That depends on your definition of slow. I'm not saying that it's a big problem,
but worth fixing considering that the fix is essentially for free. This is how I'm thinking:

The send latency (how fast the HTTP call to the SQS API completes) is typically [60-70 ms](https://softwaremill.com/amazon-sqs-performance-latency/).

The user resource (`UserResource.getUserInfoByUuid`) in the Sun has a mean response time of 6 ms.
If we would add an SQS message publishing to this, the response time would become 71 ms.

It would also mean that the resource would go from being able to handle 167 requests per second per thread, to 14 requests per second per thread.

We also have endpoints that will result in multiple messages being published.

xiaodong-izettle:
Thanks. 

> This PR is not trying to solve all scenarios, and I don't think it has to. The most common scenario, in both my current and last team, has been to send analytics messages only intended for our Growth department. These events are far from critical and a LOG.warn should be enough.

I would suggest you use whatever async solution you prefer in your service for now. You've found several quite good solutions in our repos actually. I think this library should provide a generic enough solution. :)

> The send latency (how fast the HTTP call to the SQS API completes) is typically 60-70 ms.

Are we talking about SQS or SNS? They are not the same. Possibly worse of course. :-)

> It would also mean that the resource would go from being able to handle 167 requests per second per thread, to 14 requests per second per thread.

Do you actually need to handle that many requests (167) per second per thread right now? If not, possibly no need to go for async way right now. 

> I'm not attached to this code at all and I'm not in a rush, so feel free to create another PR. My goal is just to move this forward so that we can easily do async SQS requests 

Sure, and I think we most probably are talking about SNS. :)

Thanks for the great initiative. 👍 😄 

And we do agree upon that this library needs some renovation. 😄

minisu:
Sorry for the confusion. I'm only talking about SNS. 😊 

> Do you actually need to handle that many requests (167) per second per thread right now? If not, possibly no need to go for async way right now. 

My main point was increased latency though. Amazon noticed that every 100 ms in increased response time costed them 1 % in sales (which means an even higher cost in revenue, considering fixed costs). So why would we not save 65-130 ms (in the cases where we do 1 or 2 SNS requests) if we can do it for free (in the very common case of analytics events)?

And like you have observed, most of our services already publishes to SNS async. Did they all do the wrong analysis? :-)

> I would suggest you use whatever async solution you prefer in your service for now.

Yes, that's what we'll probably end up doing.

karlgrund:
What's the status on this? I think it's a good idea and a nice approach! 👍

xiaodong-izettle:
How about something like this? https://github.com/iZettle/izettle-toolbox/pull/258

@minisu, @karlgrund and @qinjin-izettle?

PR #240, opened 2017-10-17
BugFix: truncate to earliest Offset during DST overlap

MrShish:
Background:
Instants that are before and after a dst overlap got truncated differently.
Basically instants that where during the overlap period got truncated to the earlier offset, and those
that were after, to a later offeset.
This will prevent this, and will only truncate to the earliest possible offset.


adamvoncorswant:
Looking really good. Nothing to add!

PR #241, opened 2017-10-20
Introduce empty checks for primitive array types

izrobin:
So the generic `assertNotEmpty` method doesn't work with primitive types, hence this addition.

We have a method called `boolean empty(object)` already, so let's also have this assertion method for it.

Also updated the pom to not include the dev gate repo urls

### DIFF
--- a/izettle-java/src/main/java/com/izettle/java/ValueChecks.java
+++ b/izettle-java/src/main/java/com/izettle/java/ValueChecks.java
@@ -212,6 +212,81 @@ public class ValueChecks {
     }
 
     /**
+     * Assert that an byte array has elements; that is, it must not be
+     * {@code null} and must have at least one element.
+     * <pre class="code">Assert.assertNotEmpty(array, "The array must have elements");</pre>
+     * @param array the array to check
+     * @param message the exception message to use if the assertion fails
+     * @throws IllegalArgumentException if the object array is {@code null} or has no elements
+     */
+    public static byte[] assertNotEmpty(byte[] array, String message) {
+        if (empty(array)) {
+            throw new IllegalArgumentException(message);
+        }
+        return array;
+    }
+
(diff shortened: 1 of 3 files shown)


adamvoncorswant:
Nice! What about just passing in an object instead and call the `empty`  method, or would that be a problem, considering method signature compared to the existing ones?

izrobin:
It's absolutely possible, and it doesn't collide with any other method.
I'm going to do that actually, fits better with the already existing `empty(Object)` method. Thanks!

izettle-java/src/main/java/com/izettle/java/ValueChecks.java
```diff
@@ -212,6 +212,22 @@ public static String assertNotEmpty(String text, String message) {
...
     /**
+     * Assert that an object is not "empty". See {@link ValueChecks#empty(Object)}
+     * for definition of "empty".
+     * <pre class="code">Assert.assertNotEmpty(array, "The array must have elements");</pre>
+     * <pre class="code">Assert.assertNotEmpty(optional, "The optional must exist");</pre>
+     * @param array the object to check
+     * @param message the exception message to use if the assertion fails
+     * @throws IllegalArgumentException if the object is "empty"
+     */
+    public static Object assertNotEmpty(Object array, String message) {
```
adamvoncorswant: Looking really good! One minor thing: As this method now accepts an Object, it's actually not array specific anymore, but can accept any kind of object (same as the `empty` method apparently), perhaps parameter name and javadoc should reflect that as well instead of referring to an array?

izrobin: Absolutely. I've changed it just now

adamvoncorswant:
Thanks. Looking great!

PR #244, opened 2017-10-24
Change Object to generic T for assertNotEmpty

izrobin:
I messed up in my previous PR #241 ... Of course `Object` is not going to be passed into the method, but instead a generic `<T>`. Otherwise it wouldn't be useful at all.

Sorry for the inconvenience!


karlgrund:
Makes sense! 👍

izrobin:
Thanks! 🎉

PR #246, opened 2017-11-22
Made bouncycastle dependencies optional

softarn:
Continuing with trying to get rid of the CVE errors we have.

Most users of `izettle-cryptography` aren't using org.bouncycastle dependencies and the dependencies are generating on build because of `CVE-2005-0366` and `CVE-2001-0381`. Therefore marking them as optional so they wont get added as a dependency for projects that are using `izettle-cryptography` or `izettle-messaging`. If you need to use these org.bouncycastle libraries you have to explicitly include them in your pom.

From our awesome search feature in zli it seems like it's only `sun`, `izettle-messaging` and `invoice` in which these dependencies are used.


softarn:
I'll update and add it to sun

adamvoncorswant:
Love the ambition of this! Question: when reading up on the `optional` property in maven, it seems like splitting up the project is preferred. Is that doable instead?

softarn:
The base problem here is that developers need to investigate and read up on the CVE errors for this dependency.

After some AFK talks we reached the conclusion that we'll instead try to solve this by having a global ignore file for CVE checks for issues we already know are false positives or problems that are safe to ignore.

PR #247, opened 2017-11-22
Removed netty which is unused in project

softarn:
Seems unused, should be safe to remove right?


softarn:
Works locally, doesn't build on TC. Strange

PR #248, opened 2017-11-27
Renaming and updating javadoc of AdminDirectAccessFilter->SystemDirectAccessFilter to reflect implementation

martinfridh:
Changes relate to work done in https://izettle.atlassian.net/browse/SOLID-184 as part of setting up new admin platform. Changing name/javadoc of `AdminDirectAccessFilter` will help out in clarity to not confuse usage of admin/system tokens/endpoints.


rickard-von-essen-iz:
The drawback of this is that we loos the connection to the DropWizard domain and make it more iZettle specific (the naming) :confused:.

martinfridh:
@rickard-von-essen-iz , I don't understand. Maybe because I don't fully understand the purpose of the filter to start with though 😬 

However, since the implementation is about restricting access to path `/system`, why should we name it and document it with something implying something about "Admin"?

rickard-von-essen-iz:
It restricts access to the `adminContextPath` [DropWizard configuration - Simple server](http://www.dropwizard.io/1.2.2/docs/manual/configuration.html#simple).

martinfridh:
Thanks for clarification @rickard-von-essen-iz ! Though if this is specifically related to restricting access to the dropwizard path configured for `adminContextPath` then I think we need to make the implementation either resolve that path dynamically, or at least make it possible to configure the Filter with the path used.

rickard-von-essen-iz:
Hmm it used to have that feature (_possible to configure_).

rickard-von-essen-iz:
But it would be great if it resolved it automatically (and was a proper DropWizard bundle).

martinfridh:
so, based on the described purpose of the filter I think I should close this PR. thanks for feedback @rickard-von-essen-iz

PR #249, opened 2017-12-06
Remove @author javadoc

nzroller:
The @author does not really tell who is the actual creator behind the code, `git blame` is a better option.

This is the result of running the following on Linux:

    find . -type f -name '*.java' -print0 | xargs -0 perl -0777 -pi -e 's|/\*\*[\s\*]+?\@author.+?\*/\n||sg'
    find . -type f -name '*.java' -print0 | xargs -0 perl -0777 -pi -e 's|([\s\*]+?\@author.+?\n)+|\n|sg'


adamvoncorswant:
Great!

PR #251, opened 2018-01-19
Do not panic when Time ZoneId changed between JDKs.

xiaodong-izettle:
It seems newest JDK (8_162) changed ZoneId again, probably removed `Canada/East-Saskatchewan`.

Should we do this? Or try something else? Ping @adamvoncorswant. Thanks


xiaodong-izettle:
Probably quite hard for us to change this, if we already save the integer order of this Enum in DB.

adamvoncorswant:
I think it's fine. The DB mapping is completely outside of this enum.

xiaodong-izettle:
Cool, thanks for merging. 👍

PR #252, opened 2018-01-23
Removed Canada/East-Saskatchewan from TimeZoneId

softarn:
It was removed from the JDK: http://mm.icann.org/pipermail/tz-announce/2017-October/000047.html

> Remove Canada/East-Saskatchewan from the 'backward' file, as it
exceeded the 14-character limit and was an unused misnomer anyway

Should be safe for us to remove as well since we haven't got any merchants in Canada.

### DIFF
--- a/izettle-java/src/main/java/com/izettle/java/TimeZoneId.java
+++ b/izettle-java/src/main/java/com/izettle/java/TimeZoneId.java
@@ -111,7 +111,6 @@ public enum TimeZoneId {
     AMERICA_WINNIPEG("America/Winnipeg"),
     CST6CDT("CST6CDT"),
     CANADA_CENTRAL("Canada/Central"),
-    CANADA_EAST_SASKATCHEWAN("Canada/East-Saskatchewan"),
     CANADA_SASKATCHEWAN("Canada/Saskatchewan"),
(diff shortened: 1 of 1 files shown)


xiaodong-izettle:
@adamvoncorswant, we have https://github.com/iZettle/sun/blob/master/izettle-ejb-client/src/main/java/com/izettle/ejb/entity/TimeZoneDatabaseId.java#L117. 

I don't think we are using this timezone in DB, but do we need to check in our DB to confirm?

renatoathaydes:
@xiaodong-izettle have you seen this? I think your change avoided breaking code when a new timezone was added to the JVM and we didn't map it... but when a timezone is removed, looks like a test breaks (and obviously, any user code that relies on the enum might break too).

danielefoerberghi:
New version of izettle-toolbox in artifactory?

adamvoncorswant:
I don't see a problem with this. GTG

softarn:
Had to remove this because toolbox tests were breaking the build in jdk8 162 as reported by @danielefoerberghi 

Since only the tests were breaking I don't see a need to release this. You should be able to use this lib as a dependency without a problem.

PR #255, opened 2018-02-13
Add AdminDirectAccessBundle.

### LINKED ISSUE #243: Create Bundle for AdminDirectAccessFilter
Making a Dropwizard Bundle for this would simplify usage. From this:

```java
environment.getAdminContext()
    .addFilter(AdminDirectAccessFilter.class, "*", EnumSet.allOf(DispatcherType.class));
```

...to this:

```java
bootstrap.addBundle(new AdminDirectAccessBundle())
```

Something to consider(?)

minisu:
In order to encapsulate the logic and don't hardcode paths ("/system").
A secondary goal was to not break the API of AdminDirectAccessFilter
(so existing users will not have to change their code).

Closes #243


izettle-filters/pom.xml
```diff
@@ -16,5 +16,10 @@
             <artifactId>javax.servlet-api</artifactId>
             <version>${javax.servlet-api.version}</version>
         </dependency>
+        <dependency>
+            <groupId>io.dropwizard</groupId>
+            <artifactId>dropwizard-core</artifactId>
+            <version>1.2.2</version>
```
karlgrund: Perhaps we want this as a property?

minisu:
Thanks @karlgrund. I fixed the property!

PR #256, opened 2018-02-16
izettle-messaging: Allow custom object mapper

minisu:
Without this, one cannot customize the object mapper at all (in a good way).
For example, one might want to register KotlinModule or reuse the main object
mapper of the application.

I've tried not to be too intrusive and not break the API, so formatting and architecture stays the same (for good and for worse).

The gist of the PR is that the following static factory methods were overloaded with an `ObjectMapper objectMapper` parameter:

 - `createQueueProcessor` (only for the single message type variants)
 - `encryptedMessageDispatcher`
 - `nonEncryptedMessageDispatcher`
 - `encryptedMessageQueueConsumer`
 - `nonEncryptedMessageQueueConsumer`

The old overloaded methods defaults to the `JsonSerializer.getInstance()` object mapper, to stay backwards compatible. I also renamed all variables called `jsonMapper` to `objectMapper`.


xiaodong-izettle:
Hehe, yeah. We should have `ObjectMapper` as a param in the constructor from the very beginning. 😄

xiaodong-izettle:
Let me know when you are done. 👍

minisu:
@xiaodong-izettle ok, I'm done now :-)

PR #257, opened 2018-02-22
Add support for setting attributes on messages in SQS

rillard:
This PR makes it possible to set the SQS attributes on messages. The attributes can be used to filter out messages on a more granular level than messageType.

This only adds support for filtering using SQS. Adding filtering on a programatic level would require some more work as we would like to do the filtering before we read the actual message.

An example where this would be useful is if you only want to listen to purchase messages that was created for the invoice checkout method. If we from the purchase service set an attribute called `checkout-method=invoice` the invoice-service could terraform to only listen to purchase messages that has that attribute/value combination on them - thus limiting the noise in the message handler.  

From the AWS doc:
Amazon SQS provides support for message attributes. Message attributes allow you to provide structured metadata items (such as timestamps, geospatial data, signatures, and identifiers) about the message. Message attributes are optional and separate from, but sent along with, the message body. This information can be used by the consumer of the message to help decide how to handle the message without having to first process the message body. Each message can have up to 10 attributes. To specify message attributes, you can use the AWS Management Console, AWS software development kits (SDKs), or Query API.

https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-message-attributes.html

Filtering:
https://docs.aws.amazon.com/sns/latest/dg/message-filtering.html

### DIFF
--- a/izettle-messaging/src/main/java/com/izettle/messaging/MessagePublisher.java
+++ b/izettle-messaging/src/main/java/com/izettle/messaging/MessagePublisher.java
@@ -1,8 +1,12 @@
 package com.izettle.messaging;
 
 import java.util.Collection;
+import java.util.Map;
 
 public interface MessagePublisher {
     <M> void post(M message, String eventName) throws MessagingException;
     <M> void postBatch(Collection<M> messages, String eventName) throws MessagingException;
+    default <M> void post(M message, String eventName, Map<String, String> attributes) throws MessagingException {
+        post(message, eventName);
+    }
 }
--- a/izettle-messaging/src/main/java/com/izettle/messaging/PublisherService.java
+++ b/izettle-messaging/src/main/java/com/izettle/messaging/PublisherService.java
@@ -2,28 +2,32 @@ package com.izettle.messaging;
 
 import static com.izettle.java.ValueChecks.anyEmpty;
 import static com.izettle.java.ValueChecks.empty;
+import static com.izettle.java.ValueChecks.noneNull;
 
 import com.amazonaws.services.sns.AmazonSNS;
+import com.amazonaws.services.sns.model.MessageAttributeValue;
 import com.amazonaws.services.sns.model.PublishRequest;
 import com.izettle.cryptography.CryptographyException;
 import com.izettle.messaging.serialization.DefaultMessageSerializer;
 import com.izettle.messaging.serialization.MessageSerializer;
 import java.util.Collection;
+import java.util.Map;
+import java.util.stream.Collectors;
 
 /**
  * Convenience class for using Amazon Simple Notification Service.
  */
-public class PublisherService implements MessagePublisher {
+public class PublisherService implements MessagePublisherWithAttributes {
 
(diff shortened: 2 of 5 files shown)


izettle-messaging/src/main/java/com/izettle/messaging/PublisherService.java
```diff
@@ -73,14 +77,52 @@ private PublisherService(AmazonSNS client, String topicArn, MessageSerializer me
...
     public <M> void post(M message, String eventName) throws MessagingException {
+        post(message, eventName, null);
+    }
 
+    /**
+     * Posts message to queue.
+     *
+     * @param message Message to post.
+     * @param eventName Message subject (type of message).
+     * @param attributes Attributes to be set as MessageAttributes on the publishRequest
```
martinfridh: Nice description of what the attributes are about!

martinfridh: Maybe you could describe something regarding that this implementation only supports message attributes of type `Text`, even though AWS in a sense supports some other types as well (though not really for the subscription filtering...).

rillard: We should probably allow different types as you probably want to be able to use: `String.Array`

psvanstrom:
Nice work and a really useful feature! 👍

xiaodong-izettle:
Sorry, could you please wait a bit?

xiaodong-izettle:
Added second commit, could you please have a look, @rillard? Thanks. 😄

izettle-messaging/src/main/java/com/izettle/messaging/MessagePublisher.java
```diff
@@ -1,8 +1,12 @@
 package com.izettle.messaging;
 
 import java.util.Collection;
+import java.util.Map;
 
 public interface MessagePublisher {
     <M> void post(M message, String eventName) throws MessagingException;
     <M> void postBatch(Collection<M> messages, String eventName) throws MessagingException;
+    default <M> void post(M message, String eventName, Map<String, String> attributes) throws MessagingException {
```
rillard: We talked about using a default method - but the rational behind hiding it in a sub-interface was that we don't want to affect existing code - mainly tests.

xiaodong-izettle: The existing code would not use this new method?

rillard: Ah you are correct

xiaodong-izettle: Thanks, :)

adamvoncorswant:
Nice indeed! Am I the only one that feel at least a tiny bit of reluctance, as this ties us deeper into SQS-specific functionality that's not easily found in other message buses? anyway: I'm fine with this anyway :)

PR #258, opened 2018-02-22
Add async method to MessagePublisher.

xiaodong-izettle:
I suggest we probably do something like this, if async capability is desired...


xiaodong-izettle:
Actually, I created this just after https://github.com/iZettle/izettle-toolbox/pull/234 (based on the timestamp of the commit), then I forgot. 😄

karlgrund:
Haha, nice @xiaodong-izettle 😆 
I will let @minisu have a look and see what he thinks.

PR #262, opened 2018-04-05
DoNotMerge - Making Instant JSON parsing more tolerant on time zone formatting

martinfridh:
During implementation of PaymentLink refund on iOS it turned out that our Instant JSON deserialization is not fully [RFC3339](https://www.ietf.org/rfc/rfc3339.txt) compliant, which causes problems when creating Instant strings from iOS Swift code (which, from what I understood from @jdmoreira, are including colon `:` in time zone offset). An example of Instant string that is now (ignoring this PR) not validating is `"2018-04-05T11:38:41.325+02:00"`.

Purpose of this PR is to make our Instant JSON parsing more tolerant. @adamvoncorswant @MrShish any thinking on this? Is the current (strict) implemented as such by purpose?


izettle-jackson/src/main/java/com/izettle/jackson/module/InstantRFC3339Module.java
```diff
@@ -55,20 +56,28 @@ public void serialize(
...
+                    } catch (DateTimeParseException e4) {
+                        try {
+                            return Instant.from(INSTANT_WITH_NO_MILLIS_GENERAL_TIME_ZONE_FALLBACK.parse(value));
+                        } catch (DateTimeParseException e5) {
+                            try {
+                                long timeStamp = Long.parseLong(value);
+                                return Instant.ofEpochMilli(timeStamp);
+                            } catch (NumberFormatException e3) {
+                                throw e;
+                            }
```
adamvoncorswant: This method is getting increasingly ugly (arrow style). I suppose that the below violates other principles (empty catch statements), but what about something like:
```java
            DateTimeParseException throwing;
            try {
                return Instant.from(INSTANT_WITH_ZULU_OR_OFFSET.parse(value));
            } catch (DateTimeParseException e) {
                throwing = e;
            }
            try {
                return Instant.from(INSTANT_WITH_GENERAL_TIME_ZONE_FALLBACK.parse(value));
            } catch (DateTimeParseException e2) {
            }
            try {
                return Instant.from(INSTANT_WITH_NO_MILLIS_FALLBACK.parse(value));
            } catch (DateTimeParseException e4) {
            }
            try {
                return Instant.from(INSTANT_WITH_NO_MILLIS_GENERAL_TIME_ZONE_FALLBACK.parse(value));
            } catch (DateTimeParseException e5) {
            }
            try {
                long timeStamp = Long.parseLong(value);
                return Instant.ofEpochMilli(timeStamp);
            } catch (NumberFormatException e3) {
                throw throwing;
            }

```

adamvoncorswant:
I'm ambivalent. While it's nice to be lenient for the clients, I also like our APIs to be as strict as possible. That said, we shouldn't make life impossible for our users of course...

Also, I'm really confused about the 'z' in the formatter strings, I cannot see that it matches the javadoc, but apparently it does the trick :)

Let's approve this, and leave it up to you and @MrShish to decide what to do!

xiaodong-izettle:
Could you please have a look at #263?

martinfridh:
PR replaced by https://github.com/iZettle/izettle-toolbox/pull/263

PR #264, opened 2018-04-13
Cart: Cleanup - Remove 'refund' as expression

adamvoncorswant:
As it's misleading: re*fund* is about the financial consequence of a
product return, not about the shopping cart activity.

Using the de-facto standard naming of 'return' instead.

ping @joaxand : it's at least a start :)


adamvoncorswant:
Thanks for approval, merging!

PR #266, opened 2018-04-17
Created small library to unify handling of user requests to-be-forgot…

renatoathaydes:
…ten.


adamvoncorswant:
I don't think this should be in an open repository: do you?

renatoathaydes:
This is completely abstract, simply exposing a DW task that calls "handlers" to do the actual work (which must be implemented by the services themselves), but I don't mind to move to a more appropriate repository if you have a suggestion.

PR #268, opened 2018-04-26
Cart: Better exception handling

adamvoncorswant:
When a discount is invalid, this change does so that a descriptive
IllegalArgumentException is thrown rather than a hard-to-understand
NullPointerException

ping @xiaodong-izettle


xiaodong-izettle:
Nice one 👍

PR #269, opened 2018-05-07
Fix usage of StdDateFormat.

xiaodong-izettle:
Ping @joaxand. Thanks for spotting this, and sorry for any inconveniences. 😄 👍

### DIFF
--- a/izettle-jackson/src/main/java/com/izettle/jackson/module/InstantRFC3339Module.java
+++ b/izettle-jackson/src/main/java/com/izettle/jackson/module/InstantRFC3339Module.java
@@ -25,7 +25,6 @@ import java.time.Instant;
  */
 public class InstantRFC3339Module extends SimpleModule {
 
-    private static final StdDateFormat STD_DATE_FORMAT = new StdDateFormat();
     /**
      * Note: This module needs to be registered after other possible modules that might try to control `Instant`, such
      * as the JaveTimeModule.
@@ -56,7 +55,7 @@ public class InstantRFC3339Module extends SimpleModule {
         ) throws IOException {
             final String value = jp.readValueAs(String.class);
             try {
-                return STD_DATE_FORMAT.parse(value).toInstant();
+                return new StdDateFormat().parse(value).toInstant();
             } catch (ParseException e) {
                 throw new IOException(e);
             }


izettle-jackson/src/main/java/com/izettle/jackson/module/InstantRFC3339Module.java
```diff
@@ -56,7 +55,7 @@ public Instant deserialize(
         ) throws IOException {
             final String value = jp.readValueAs(String.class);
             try {
-                return STD_DATE_FORMAT.parse(value).toInstant();
+                return StdDateFormat.instance.clone().parse(value).toInstant();
```
adamvoncorswant: Why call the obscure `clone` method (which normally raises a few eyebrows), instead of just doing `new StdDateFormat()` here instead?

xiaodong-izettle: Well, it's not obscure to me as it's overridden exactly in `StdDateFormat`. 😄 

https://github.com/FasterXML/jackson-databind/blob/master/src/main/java/com/fasterxml/jackson/databind/util/StdDateFormat.java#L241-L246

And this is how Jackson use this object 

https://github.com/FasterXML/jackson-databind/blob/master/src/main/java/com/fasterxml/jackson/databind/DeserializationContext.java#L1638-L1649

adamvoncorswant: Yes, I noticed that as well. But the point was just that I think using a `clone` method adds cognitive load to the reader of the code (at least I will always go in and check the source). What's the value of calling `clone` instead of just new:ing?

joaxand: I'm also voting for new. Then we don't have to worry about the ugly setters on StdDateFormat.instance either.

xiaodong-izettle: I think the library user could do all kinds of optimisations further on for `clone`, but much harder for the constructor.

joaxand: True, but it doesn't matter how optimized it is when someone calls `StdDateFormat.instance.setTimeZone(null);`.

adamvoncorswant: Ok, I'm not going to fight you on this so go with it if you feel it's good as is. Just one last word:

I'm not aware of *any* other place in our codebases where we create one phony instance that we call `clone` upon, just to create new identical objects... I don't see the possible optimizations here either, as we in the end still need to have a *new* object anyway. The standard practice here would be to have factory methods, or to simply call `new`.

xiaodong-izettle: May I ask how you tried this, @joaxand? 

> True, but it doesn't matter how optimized it is when someone calls StdDateFormat.instance.setTimeZone(null);.

I hope there is good reason for someone doing that. 😄 

Also @adamvoncorswant, sorry for the misunderstanding first. By `I think the library user could do all kinds of optimisations further on for clone`, I actually meant `I think the library author could do all kinds of optimisations further on for clone`. 😄 Such a lethal typo....

I do not have that strong opinion on this either. I just would like to follow how the library author uses this method. I think the author should have some reason here. I'll do a small micro-benchmark later, to see if there is any performance gain, that could justify using `clone` here, @adamvoncorswant. 

> I'm not aware of any other place in our codebases where we create one phony instance that we call clone upon, just to create new identical objects

2 examples: 

https://github.com/iZettle/sun/blob/master/izettle-ejb/src/main/java/com/izettle/ejb/bean/UserFacade.java#L535-L536

https://github.com/iZettle/sun/blob/master/izettle-ejb/src/main/java/com/izettle/ejb/bean/cashregister/CashRegisterApplicationManager.java#L856

And there are a few in `crypto` and elsewhere. 🙂

adamvoncorswant: Haha, but those objects are not *phony* ones, created with an empty constructor...

adamvoncorswant:
Am I a bore if I suggest that we revert back to the java8-ish kind of solution, and just add a new allowed format instead? java.time FTW!

izettle-jackson/src/main/java/com/izettle/jackson/module/InstantRFC3339Module.java
joaxand: @xiaodong-izettle 
Can't reproduce the original issue again with these changes now. Sorry about the last comment, withdrawn :)

xiaodong-izettle: Cool, thanks for the info, @joaxand. 

I just run some micro-benchmark with both `clone()` and the default constructor. No obvious difference... 

So I changed the `clone()` usage...

PR #271, opened 2018-05-09
[WIP] GDPR event handling

danielefoerberghi:
Common RBTF GDPR consumer code, to reduce the amount of duplication of code in each service.
The responsibility for the services are to implement a GDPR Dao interface. And implement a QueProcessor which uses the the general GdprManager from this package. Some functions in the GdprManager can be overidden. The endpoint in the services also need to create a MessagePublisher used to publish an GdprEvent on the specific queue for on-demand requests. In the case hops publishes events uses the same flow.


adamvoncorswant:
Do we really feel that this code is completely void of iZettle-specific logic and deserves a place in an open-source repo?

ericsjogren:
@adamvoncorswant 
Oh at least I was not aware that this was an open-source library, must have missed that.
Think I also got confused by the line: `Libraries built and used internally at iZettle.`

Do you have a recommendation for a better place to put this?

psvanstrom:
Perhaps https://github.com/iZettle/izettle-commons would be a better repo? We could also put this into its own repo, a small streamlined GDPR-library.

adamvoncorswant:
I'm closing this. Taking the discussion out of public forums.

PR #272, opened 2018-05-11
Introduces izettle-kotlin library

olsson:
This PR introduces a new `izettle-kotlin` library into the toolbox. The motivation for this library is simply to make programming in Kotlin at iZettle easier and more enjoyable, and to reduce boilerplate and ceremony (a primary goal of Kotlin)

Initially, I have provided some functions that I think will make our Kotlin code more "Kotlin-like", by encouraging chains of calls and not `if` statements. We are already using the `whenNotNull` function in https://github.com/iZettle/online-payment and it would be much better to have it in a library like this one.

Here is a summary of the features this PR will provide:

```kotlin
// Introduces a whenNoneNull() global function that acts like a multi-variable ?.let()
// Supports between 2 and 9 arguments of different or same types

whenNoneNull(nullableName, nullableAge) { name, age ->
    println("Block was called with $name, $age, guaranteed to be non null")
} ?: throw Exception()


// Adds a mapper between Java Optional to a nullable Kotlin type, 
// to cut down on boilerplate in Kotlin code where we get Optionals 
// from legacy Java frameworks or code

val result = optional.asNullable()?.let { "has value $it" } ?: "empty"
```

### DIFF
--- a/izettle-kotlin/pom.xml
+++ b/izettle-kotlin/pom.xml
@@ -1,0 +1,89 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <groupId>com.izettle.toolbox</groupId>
+        <artifactId>izettle-toolbox</artifactId>
+        <version>1.0.122-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>izettle-kotlin</artifactId>
+    <version>1.0.122-SNAPSHOT</version>
+
+    <properties>
+        <kotlin.version>1.2.71</kotlin.version>
+    </properties>
+
+    <dependencies>
+
+        <dependency>
+            <groupId>com.izettle.toolbox</groupId>
+            <artifactId>izettle-java</artifactId>
+            <version>1.0.122-SNAPSHOT</version>
+        </dependency>
+
+        <dependency>
+            <groupId>org.jetbrains.kotlin</groupId>
+            <artifactId>kotlin-stdlib</artifactId>
+            <version>${kotlin.version}</version>
+            <scope>provided</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.jetbrains.kotlin</groupId>
+            <artifactId>kotlin-reflect</artifactId>
+            <version>${kotlin.version}</version>
+            <scope>compile</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.jetbrains.kotlin</groupId>
+            <artifactId>kotlin-test-junit</artifactId>
+            <version>${kotlin.version}</version>
+            <scope>test</scope>
+        </dependency>
+
+    </dependencies>
+
+    <build>
+        <sourceDirectory>${project.basedir}/src/main/kotlin</sourceDirectory>
+        <testSourceDirectory>${project.basedir}/src/test/kotlin</testSourceDirectory>
+        <plugins>
+            <plugin>
+                <artifactId>kotlin-maven-plugin</artifactId>
+                <groupId>org.jetbrains.kotlin</groupId>
+                <version>${kotlin.version}</version>
+
+                <executions>
+                    <execution>
+                        <id>compile</id>
+                        <phase>process-sources</phase>
+                        <goals>
+                            <goal>compile</goal>
+                        </goals>
+                    </execution>
+
+                    <execution>
+                        <id>test-compile</id>
+                        <phase>process-test-sources</phase>
+                        <goals>
+                            <goal>test-compile</goal>
+                        </goals>
+                    </execution>
+                </executions>
+            </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-compiler-plugin</artifactId>
+                <inherited>true</inherited>
+                <configuration>
+                    <showDeprecation>true</showDeprecation>
+                    <showWarnings>true</showWarnings>
+                    <optimize>true</optimize>
+                </configuration>
+            </plugin>
+        </plugins>
+    </build>
+
+</project>
--- a/izettle-kotlin/src/main/kotlin/com/izettle/kotlin/Optionals.kt
+++ b/izettle-kotlin/src/main/kotlin/com/izettle/kotlin/Optionals.kt
@@ -1,0 +1,15 @@
+package com.izettle.kotlin
+
+import java.util.Optional
+
+/**
+ * Convert a Java Optional to a Kotlin nullable type to avoid handling Optionals in Kotlin.
+ *
+ * Usage:
+ * ```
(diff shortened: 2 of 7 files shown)


izettle-kotlin/src/main/kotlin/com/izettle/kotlin/Optionals.kt
```diff
@@ -0,0 +1,15 @@
...
+ * Convert a Java Optional to a Kotlin nullable type
+ *
+ * Usage:
+ * ```
+ * optional.asNullable()?.let { "has value $it" } ?: "empty"
+ * ```
+ *
+ * @see Optional
+ */
+fun <T : Any> Optional<T>.asNullable(): T? = if (this.isPresent) this.get() else null
```
pvoid: why not just `orElse(null)` instead of if statement?

izettle-kotlin/src/main/kotlin/com/izettle/kotlin/ValueChecks.kt
```diff
@@ -0,0 +1,164 @@
...
+ * Return the result of the provided block if the value is empty according to ValueChecks.empty()
+ *
+ * Usage:
+ * ```
+ * val result = field.letIfEmpty { "empty" } ?: "the $field was not empty"
+ * ```
+ *
+ * @see com.izettle.java.ValueChecks.empty
+ */
+fun <T : Any?, R : Any?> T?.letIfEmpty(block: () -> R?): R? = if (ValueChecks.empty(this)) block() else null
```
pvoid: Is it just a replacement for 

`field.takeIf { ValueChecks.empty(this) }?.let { block() }`

Does it worth it?

izettle-kotlin/src/main/kotlin/com/izettle/kotlin/Optionals.kt
olsson: True. Simpler

izettle-kotlin/src/main/kotlin/com/izettle/kotlin/ValueChecks.kt
olsson: Yes, you could definitely solve it like that too!

There's a lot of `if (!empty(field)) { block() } else { null }` scattered around since we are all so used to using `ValueChecks.empty` like that. This would maybe encourage us to move forward past that syntax in Kotlin.

izettle-kotlin/src/main/kotlin/com/izettle/kotlin/ValueChecks.kt
```diff
@@ -0,0 +1,164 @@
...
+ * ```
+ * val field1: String? = null
+ * val field2: Int? = null
+ * whenNoneNull(field1, field2) { f1, f2 ->
+ *    f1.toUppercase() // non-nullable f1
+ *    f2.toString()    // non-nullable f2
+ * }
+ * ```
+ */
+fun <T1 : Any, T2 : Any, R : Any> whenNoneNull(p1: T1?, p2: T2?, block: (T1, T2) -> R?): R? {
```
pvoid: My advice - mark generic types as `in`. Can prevent some future problems.

pvoid: Actually, it is a common advice. As for me, generics in Kotlin are slightly ... let's call it messy. It is always better to specify intention for generic in or out. Otherwise, users can struggle with weird compile complains.

olsson: It doesn't seem like the covariance modifiers work on generic type parameters in global extension function, only on functions that are part of classes or interfaces. Am I wrong? How would it look?

pvoid: Yep, you right. My bad, sorry.

izettle-kotlin/src/main/kotlin/com/izettle/kotlin/ValueChecks.kt
pvoid: Well, ok, as you wish. Then it smth personal, as for me in my example code is more obvious without navigating inside library function.

Then an advice maybe it worth to add takeIfEmpty extension

izettle-kotlin/src/main/kotlin/com/izettle/kotlin/ValueChecks.kt
```diff
@@ -0,0 +1,171 @@
...
+ *
+ * Usage:
+ * ```
+ * val result = field.letIfEmpty { "empty" } ?: "the $field was not empty"
+ * ```
+ *
+ * @see com.izettle.java.ValueChecks.empty
+ */
+inline fun <T : Any?, R : Any?> T?.letIfEmpty(block: () -> R?): R? =
+    if (ValueChecks.empty(this)) block() else null
```
minisu: I get that it's handy to be able to do use this function together with `?:` (as in the KDoc example), but isn't it odd that this function returns null?

In the example, `?:` is used semantically like `else`, while it's real meaning is strictly related to nullability (just like the other operators and type declarations containing a `?`).

An alternative could be a function called `emptyToNull`, used like this:

```kotlin
field.emptyToNull?.let { "empty" } ?: "the $field was not empty"
```

Side note: I think `empty` is overused (which is a bit dangerous) at iZettle.

olsson: Maybe it wasn't super clear, but my thinking was that Kotlin's built-in `.let { }` applies the provided lambda _if not null_, and returns null if it was null. So similarly `.letIfNotEmpty { }` would follow the same pattern as `.let { }` but apply the lambda _if not empty_ and null if empty. The naming `.let*` should signal the similar behavior. 

I do think `null` is an appropriate response type. If you take the expression:
```kotlin
field.filter { ValueChecks.empty(it) }?.let { "empty" } ?: "$field is not empty"
```

If you think of the part of the expression `field.filterNot { ValueChecks.empty(it) }`. It will return `null` if `field` is empty and this doesn't really seem that strange to me. What do you think?

olsson: We could also just ditch the `.letIfEmpty` functions, since the `whenNoneNull` is what I would most like to get into this library.

izettle-kotlin/pom.xml
```diff
@@ -0,0 +1,146 @@
...
+        <version>1.0.117-SNAPSHOT</version>
+    </parent>
+
+    <groupId>com.izettle.toolbox</groupId>
+    <artifactId>izettle-kotlin</artifactId>
+    <version>1.0.117-SNAPSHOT</version>
+
+    <properties>
+        <kotlin.version>1.2.41</kotlin.version>
+        <ktlint.version>0.22.0</ktlint.version>
```
magnuslambert: Maybe update to `0.23.1`

izettle-kotlin/src/main/kotlin/com/izettle/kotlin/ValueChecks.kt
olsson: Hm good point. I will just get rid of these functions. Any comments on the rest?

minisu:
@olsson wrote:
> Hm good point. I will just get rid of these functions. Any comments on the rest?

The rest I like :+1:

olsson:
@aaanders updated basepom and rebased. care to review?

PR #274, opened 2018-05-30
In order to enable Cassandra Auditing.

xiaodong-izettle:
Removing `astyanax` stuff that we have moved away from. 

Add configuration for Cassandra system user `username` and `password`.


xiaodong-izettle:
Ping @x1a0, as you are the original author? 

And also F.Y.I @progre55, we need to for Purchase service and Cash Register service.

xiaodong-izettle:
Thanks, @x1a0. 😄

PR #277, opened 2018-08-21
Improve Cassandra SchemaVersionUpdaterWithDatastaxDriver.

xiaodong-izettle:
So that we could reliably use it in ECS env. 

Ping @Sarastro72, thanks.


xiaodong-izettle:
Or ping @x1a0. Thanks.

xiaodong-izettle:
Thanks. Merging...

PR #278, opened 2018-08-29
Enforce that adminContextPath is set to /system

minisu:
- If set to /system: Do nothing
 - If set to something else: Throw
 - If not set at all: Set it to /system

This removes the need for any config changes for services --
all they need is the bundle. This is an enabler towards having
an iZettle Dropwizard flavor, with all necessary config,
that works OOTB.


nzroller:
:+1:

PR #279, opened 2018-09-03
Add back missing method, as Sun is still using it.

xiaodong-izettle:
https://github.com/iZettle/sun/blob/master/izettle-ejb/src/main/java/com/izettle/ejb/general/JmsMessageQueueConsumer.java#L48

Ping @minisu, thanks.


izettle-messaging/src/main/java/com/izettle/messaging/serialization/MessageDeserializer.java
```diff
@@ -51,8 +51,12 @@ public M deserialize(String message) throws IOException {
     }
 
     public String removeSnsEnvelope(String message) throws IOException {
+        return removeSnsEnvelope(objectMapper, message);
+    }
+
+    public static String removeSnsEnvelope(ObjectMapper om, String message) throws IOException {
```
minisu: Hmm, the only usage I can find in the Sun is [this](https://github.com/iZettle/sun/search?q=removeSnsEnvelope&unscoped_q=removeSnsEnvelope); but that's using `removeSnsEnvelope(String)`, not `removeSnsEnvelope(ObjectMapper, String)`. Maybe it's better to make that class in the Sun (`JmsMessageQueueConsumer`) have a `MessageDeserializer` in a field?

xiaodong-izettle: Yeah, but it has a `ObjectMapper` in scope that could be easily passed into `removeSnsEnvelope(ObjectMapper, String)` here: https://github.com/iZettle/sun/blob/master/izettle-ejb/src/main/java/com/izettle/ejb/general/JmsMessageQueueConsumer.java#L20, yes?

xiaodong-izettle: Ping @minisu.

minisu: Aha, ok :+1:

PR #281, opened 2018-09-27
Add missing `findsbugs/excludes.xml`

AndersKOlsson:
- So `static-analysis` profile works


AndersKOlsson:
So, 20 bugs flagged. Should this CI job really fail just on possible bugs? There is an error classification, as well.

xiaodong-izettle:
Thanks for the effort. 👍 😄

AndersKOlsson:
Feels like it could be a long and windy road, tho...

AndersKOlsson:
Do you gentlemen think we should go through, and clean them up?

AndersKOlsson:
Now we're down to `18` now. IMHO, the rest could be excluded. Do you agree?

AndersKOlsson:
I've fixed and whitelisted for 3 modules now. But the `izettle-messaging` module has so many, and I can't justify putting more time on it.

xiaodong-izettle:
Should be fine now. Feel free to merge.

izettle-java/src/main/java/com/izettle/java/ArrayUtils.java
```diff
@@ -9,11 +8,11 @@ private ArrayUtils() {
     }
 
     public static byte[] concat() {
-        return new byte[]{};
+        return null;
     }
 
     public static byte[] concat(byte[] a) {
-        return Arrays.copyOf(a, a.length);
+        return a;
```
AndersKOlsson: Did you think this change would break something? If not, I think it's not a correct implementation to return the same array, since it's a mutable data type, and the rest of the methods return copies, which are safe to/from modifying.

xiaodong-izettle: Probably it could break something, it's really old code. And it could have issues when the byte array a is REALLY big... 

Let's stick with the old code for now, 🙂

AndersKOlsson:
I restaged the commits a bit. Content remains as is, tho.

PR #285, opened 2018-11-13
Java toolbox: Java 11 adoptions

adamvoncorswant:
Turns out that the currency format for different locales has changed in
newer versions of java. Most significantly, non breaking space and
unicode "minus sign" is sometimes used instead of horizontal space and
hyphen.

We choose to live with this change, as we don't want to add custom
configuration that's costly to maintain. Also, given that other repos already use this artifact, they are already unknowingly formatting in the new way if they're running on java11.

But, as a consequence of this, the parser cannot parse strings with
normal space or hyphen, should those be the characters preferred by the
formatter. So, to not make this too dangerous, logic has been changed in
the parser method, so that normal space and hyphen won't cause a
problem.

ping @minisu and @xiaodong-izettle 

Unfortunately, the entire repo doesn't build under my java11 installation, as some dependency in izettle-cassandra-datastax seems to break under java11. Any idea on how to solve that?


adamvoncorswant:
@minisu or @xiaodong-izettle : any thoughts on this?

adamvoncorswant:
Thanks @minisu ! Still have to solve the build failure though...

pom.xml
```diff
@@ -92,6 +92,15 @@
     </dependencyManagement>
     <build>
         <plugins>
+            <plugin>
+                <artifactId>maven-compiler-plugin</artifactId>
+                <configuration>
+                    <compilerArgs>
+                        <arg>-parameters</arg>
+                    </compilerArgs>
+                    <release>11</release>
```
adamvoncorswant: Nice! So we upgrade the entire toolbox to java 11 (and not just izettle-java)?

xiaodong-izettle: `izettle-java` is used by almost every repo, yes? If we could bump that to java-11, probably we could the entire toolbox to java-11?

adamvoncorswant: I agree. You're right!

adamvoncorswant:
Great! I still get: 
```
[ERROR] useConfiguredValues(com.izettle.cassandra.CassandraSessionFactoryTest)  Time elapsed: 0.001 s  <<< ERROR!
java.lang.NoClassDefFoundError: Could not initialize class org.apache.cassandra.io.util.FileUtils                             
        at com.izettle.cassandra.CassandraSessionFactoryTest.before(CassandraSessionFactoryTest.java:29)                      

```
but it seems our build server is happy anyway.

adamvoncorswant:
Thank you for help with this @xiaodong-izettle !

PR #287, opened 2018-11-26
Enforce even length on hex when converting to byte[]

FredrikMunter:
* Don't allow odd length input when doing hex to byte array conversion. It currently drops the last character silently because of [this](https://github.com/iZettle/izettle-toolbox/blob/4a9733765b4706a40bee3c1867e7591092512060/izettle-java/src/main/java/com/izettle/java/Hex.java#L38) conversion.
* Remove the identical `Hex` class from module `izettle-emv` - it's dependent on `izettle-java` which has the same class anyway.
* Add some test cases for weird input just to make sure


adamvoncorswant:
:+1:

